PR #41, opened 2022-06-06
IGVF-97 add edit json

dougli1sqrd:
This adds a JSON editor for database objects in the web UI.

Each object page gets an "Edit JSON" button if you are logged in and have edit privileges on that object. 

Clicking the button brings you to the edit page which has a text field with the current JSON. If any JSON errors are present when typing the "Save" button is disabled and a small indicator will show that there is a syntax error. Mousing over will provide info what went wrong.

Clicking save will attempt to `PATCH` the database object with the contents of the JSON field. If successful, the editor closes returning you to the database object page. Otherwise if an error occurs the editor page will stay and you can continue to edit. An error message shows at the bottom indicating something went wrong during the save. Current this error message is fairly generic as I'm having a hard time reaching the error Response object which actually contains the error description.

Lastly, if someone navigates directly to the page (via address bar) and they do not have proper privileges then the JSON will appear (with whatever fields the backend deems appropriate for that priv level) and neither editing (typing in the field) nor saving are enabled.


components/button.js
```diff
@@ -71,25 +74,28 @@ const buttonSizeClasses = {
...
+export const Button = React.forwardRef(({
   onClick,
   type = "primary",
   size = "md",
   label = "",
   className = "",
+  enabled = true,
   children,
-}) => {
+}, ref) => {
```
forresttanaka, line 85: How does the ref property work? I’m not familiar with it used this way, and I couldn’t find anything relevant in the React documentation.

forresttanaka (changes requested):
It works really well! I have some suggestions, and I’d really like to enforce commenting for reasons I give in my review comments (tl;dr it helps me be fast and lazy). Also generally I think you can remove the `path` property from `<EditLink>`, which will simplify all those calls. I’ll also put some styling suggestions in the ticket, though you should get feedback from people who’d actually use it. Meenakshi especially has lots of opinions on UI, so I often look to her for that. You can also bring it up with Ben for discussion in a group or project meeting if he thinks its worth it.

forresttanaka:
Some more thoughts on why I’d like to enforce commenting the code more if it doesn’t make me more enemies. I’m a lazy code reviewer and want to get it out of the way quickly, and non-commented code slows me down, which in turn puts more pressure on QA since they get less time. I also like to use the JSDoc form so that when you hover over a function call, a pop-up appears showing a description of the function as well as its parameters. Like, when I hovered over `canEdit`, nothing of value appeared and I had to go to that function instead of being able to read what it’s for and what its parameter is for (though admittedly in that case, it’s pretty obvious — it’s often far from obvious). In addition, encoded has code that we don’t know why it’s there because it doesn’t make sense and has no comment, and the person who wrote it couldn’t remember, so now we’re just stuck with code we might or might not need.

forresttanaka:
On PR naming, you’ll find people do it different ways, but the name starting with IGVF-nn is basically a requirement, though the default is “Igvf nn” which Idan doesn’t seem to have an issue with. In encoded, PR names get used in the public release notes, so I like to use “ENCD-nn Fixed something and added something” which I find helpful as a consumer of the release notes. But Keenan for example likes “ENCD-nn-fixed-something-and-added-something” which I find more difficult to read, but I doubt you’ll find many other than me who care.

components/edit-func.js
```diff
@@ -0,0 +1,133 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import dynamic from 'next/dynamic'
+
+import { LinkButton } from './button'
+import { canEdit } from '../libs/general'
+
+/*eslint no-unused-vars: ["error", { "varsIgnorePattern": "^_.*" }]*/
```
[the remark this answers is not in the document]
dougli1sqrd: This fixes unused variables in `filterItem`:
```
const filterItem = (item) => {
    const {
        ['@type']: _type,
        ['@id']: _id,
        ['@context']: _context,
        ['uuid']: _uuid,
        ['actions']: _actions,
        ['access_keys']: _access,
        ['schema_version']: _version,
        ['title']: _title,
        ...filtered } = item
    return filtered
}
```

I don't know, it was giving me trouble, so I got it to ignore unused variable lint errors on vars prepended with underscore.

pages/[type]/[identifier]/edit.js
```diff
@@ -0,0 +1,210 @@
...
+    return (
+        <div>
+            <p className="text-rose-600 text-lg">Errors from Save:</p>
+            <ul className="list-disc">
+                {errors.map((error) =>
+                    <li key={error.description} className="text-rose-600 text-base">
+                        {error.description}
+                    </li>
+                )}
+            </ul>
```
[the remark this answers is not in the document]
Comment (dougli1sqrd) [reply pages/[type]/[identifier]/edit.js:91]: Oh this `SavedErrors` component is for if the formatting is wrong on the server, or some other random thing happens. Ideally the server will come back with an error response that has a description of went wrong and we can display it. Since it could be a formatting issue: for example if it's supposed to be  CURIE but a curie is not supplied the server responds with a 415 or 422 or something.

pages/[type]/[identifier]/edit.js
```diff
@@ -0,0 +1,210 @@
...
+        method: "PATCH",
+        credentials: "include",
+        headers: {
+            "Content-Type": "application/json",
+            "X-CSRF-Token": session._csrft_,
+        },
+        body: JSON.stringify(item),
+    }).then(response => {
+        return response
+    }).catch(_err => {
```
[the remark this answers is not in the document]
Comment (dougli1sqrd) [reply pages/[type]/[identifier]/edit.js:25]: whoa you can just delete it! Wild!

pages/[type]/[identifier]/edit.js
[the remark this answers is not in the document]
Comment (dougli1sqrd) [reply pages/[type]/[identifier]/edit.js:91]: I guess we could also try at some point is using the same template validation as on the backend and show those errors also? Could be a cool feature at some point.

components/button.js
dougli1sqrd: To be honest I don't totally understand, but this is something I found. I found it associated with link buttons?
https://reactjs.org/docs/forwarding-refs.html

We also use it in `NavigationButton` in navigation.js. I imagine I was copying from there too.

pages/[type]/[identifier]/edit.js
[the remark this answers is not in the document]
Comment (forresttanaka) [reply pages/[type]/[identifier]/edit.js:25]: Well, replace `_err` with `()` will get rid of the error.

components/edit-func.js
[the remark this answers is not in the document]
forresttanaka: Oh, I see the error in edit.js. But it’s not needed in _this_ file, is it?

pages/[type]/[identifier]/edit.js
```diff
@@ -0,0 +1,210 @@
...
+        ['@id']: _id,
+        ['@context']: _context,
+        ['uuid']: _uuid,
+        ['actions']: _actions,
+        ['access_keys']: _access,
+        ['schema_version']: _version,
+        ['title']: _title,
+        ...filtered } = item
+    return filtered
+}
```
Comment (forresttanaka) [inline pages/[type]/[identifier]/edit.js:65]: Oh, I missed this before. I think this is what you asked about before, where you wanted to copy an object but with certain properties filtered out. I think you should do a fetch of the object with ?frame=edit so that the back end can determine what it does and doesn’t want in a writeable object, instead of filtering an object on the front end.

pages/[type]/[identifier]/edit.js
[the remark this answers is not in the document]
Comment (forresttanaka) [reply pages/[type]/[identifier]/edit.js:91]: On validation, I’d love to see some standard way, possibly shared across back end and front end, to show (or I suppose in some cases hide) errors to the user in a consistent way. It could be a modal, or maybe a small message in some consistent area — I don’t know.

components/edit-func.js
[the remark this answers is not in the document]
forresttanaka: Oh, dang, I remember now. I completely forgot that _I_ copied that code. Yeah, the ref contains a reference to the button DOM element. It’s puzzling-looking here because we don’t actually see ref used, but the `<Link>` component uses it internally. So yeah, you have it right.

components/button.js
forresttanaka: Oh, dang, I remember now. I completely forgot that I copied that code. Yeah, the ref contains a reference to the button DOM element. It’s puzzling-looking here because we don’t actually see ref used, but the <Link> component uses it internally. So yeah, you have it right.

pages/[type]/[identifier]/edit.js
Comment (dougli1sqrd) [reply pages/[type]/[identifier]/edit.js:83]: oh yes I can change that as well then

forresttanaka (changes requested):
I think this looks a lot more maintainable. Make sure you install prettier or the next person to edit these files will have a ton of changes they have nothing to do with. I think the `useEditor` hook works well, but it does make me wonder if a wrapper component would be nicer for pages to use, like `<EditorEnvironment>{page stuff}</EditorEnvironment>` instead of `{useEditor()`. I’m not sure that’s better, but I think it’d be more straightforward for people making more pages in the future to use a wrapper component. It’d be nice to see more consistent comment usage.

forresttanaka:
Another thing to think about is adding a Cypress test that edits an object and confirms the object gets changed appropriately. There might be some Jest tests you can add too, though Jest tests are by design a real commitment. If you add a Jest test to a file, the coveralls module will force you to cover every line in the tested file, and every condition in every relevant line. That’s another reason to try to keep file sizes small — to avoid having to commit to too much because of a Jest test.

libs/sessionFetch.js
```diff
@@ -0,0 +1,77 @@
...
+          {
+            description: "Network error while updating",
+            names: ["unknown"],
+          },
+        ],
+      }
+    })
+    return response
+  }
+}
```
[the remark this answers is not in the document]
dougli1sqrd: I was thinking that the getObject method could also be HEAD.

Ohh I should `await fetch` like in request?

components/edit.js
```diff
@@ -0,0 +1,253 @@
...
+
+SaveCancelControl.propTypes = {
+  cancelClick: PropTypes.func.isRequired,
+  saveClick: PropTypes.func.isRequired,
+  itemPath: PropTypes.string.isRequired,
+  saveEnabled: PropTypes.bool.isRequired,
+}
+
+function sortedJson(obj) {
+  if (obj instanceof Array) {
```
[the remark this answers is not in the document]
forresttanaka: Yeah, encoded didn’t do this the best way it seems. That encoded code is really old, and I’m not sure if the recommendation was different back then.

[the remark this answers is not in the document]
dougli1sqrd: Hmm when I do that eslint says it's an invalid comparison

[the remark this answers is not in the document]
forresttanaka: `typeof obj === "object"`?

libs/sessionFetch.js
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  async getObject(path, method) {
+    const response = await fetch(`${API_URL}${path}`, {
+      method: method,
+      credentials: "include",
+      headers: buildHeaders(this.session, false),
+    }).then((response) => {
+      return response
+    })
```
forresttanaka, line 64: The .then() clause is redundant since you await fetch. You can surround the whole thing with a try/catch instead of chaining the catch to a then().

libs/sessionFetch.js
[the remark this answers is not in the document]
forresttanaka: I feel like someday we’ll need to combine this with what’s already in request.js. Right now I’m not sure whether I’m supposed to use this Fetch class to get an object, or if I should use the the `Request` class’s `getObject()` in request.js. Seems like a lot of duplicate functionality.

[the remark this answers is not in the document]
forresttanaka: I’m hoping not to fall into the same mix of shishkebab, snake, and camel case in file names like we have in encoded. [NextJS itself uses shishkebab case](https://github.com/vercel/next.js/tree/canary/examples) so that’s why I chose that for igvf-ui. That’s also why I chose indenting by 2 spaces, and no semicolons — that’s all from NextJS.

components/edit.js
```diff
@@ -0,0 +1,235 @@
...
+  <>
+    {viewComponent}
+    {editlink}
+  </>
+  return componentToShow
+}
+
+export const EditableItem = ({ item, children }) => {
+  return useEditor(item, children)
+}
```
forresttanaka, line 48: It’s so much nicer to use `EditableItem` on the pages now. Is there any value to having the `useEditor` custom hook though? That’s useful when you need the same React code shared across multiple components, but that’s not happening here. Is that likely to happen in the future, or can we just put all of `useEditor` within `EditableItem`?

components/edit.js
```diff
@@ -0,0 +1,235 @@
...
+
+  const [editorStatus, setEditorStatus] = useState({
+    canEdit: editable(session, item),
+    canSave: editable(session, item),
+    errors: [],
+  })
+
+  useEffect(() => {
+    const fetch = new Fetch(session)
+    fetch.getObject(`${path}?frame=edit`, "GET").then((value) => {
```
forresttanaka, line 157: I think it’s confusing that you have to pass the "GET" operation to a method called `getObject()`. Would you ever pass "PUT" to `getObject()`?

components/edit.js
```diff
@@ -0,0 +1,235 @@
...
+    }
+  }
+
+  const { session } = useContext(SessionContext)
+
+  const [text, setText] = useState(() => {
+    JSON.stringify({}, null, 4)
+  })
+
+  const [saveErrors, setErrors] = useState([])
```
forresttanaka, line 147: Unusual naming for a state. Usually you see `[errors, setErrors]`, like you have below with `editorStatus`.

components/edit.js
```diff
@@ -0,0 +1,235 @@
...
+
+  const [editorStatus, setEditorStatus] = useState({
+    canEdit: editable(session, item),
+    canSave: editable(session, item),
+    errors: [],
+  })
```
forresttanaka, line 153: By initializing a state with an object that calls functions, those functions get called at every render, but their results get thrown away after the first render.

components/edit.js
```diff
@@ -0,0 +1,235 @@
...
+SavedErrors.propTypes = {
+  errors: PropTypes.array.isRequired,
+}
+
+const EditPage = ({ item }) => {
+  const path = item["@id"]
+
+  const loggedIn = useAuthenticated()
+
+  const editable = (session, item) => {
```
forresttanaka, line 124: The `session` argument never gets used.

components/edit.js
```diff
@@ -0,0 +1,235 @@
...
+  }, [path, session])
+
+  const router = useRouter()
+
+  const onChange = (newValue) => {
+    setText(newValue)
+    const errors = jsonErrors(newValue)
+    const status = {
+      canEdit: editable(session, item),
+      canSave: errors.length == 0 && editable(session, item),
```
forresttanaka, line 172: Calling the same function with the same arguments twice.

components/edit.js
```diff
@@ -0,0 +1,235 @@
...
+          canSave: true,
+          errors: [],
+        })
+        const errors = response.errors.map((err) => ({
+          description: err.description,
+          keys: err.names
+            .map((val) => {
+              `\`${val}\``
+            })
+            .join(", "),
```
forresttanaka, line 203: The .map() callback doesn’t return a value.

components/edit.js
```diff
@@ -0,0 +1,235 @@
...
+      </div>
+    </div>
+  )
+}
+
+EditPage.propTypes = {
+  item: PropTypes.object.isRequired,
+}
+
+export default EditPage
```
forresttanaka, line 235: This file isn’t prettier compliant. The .then() formatting changes when you save the file.

components/edit-func.js
```diff
@@ -0,0 +1,133 @@
...
+        return null
+    }
+}
+
+EditLink.propTypes = {
+    path: PropTypes.string.isRequired,
+    item: PropTypes.object.isRequired,
+}
+
+export default EditJson
```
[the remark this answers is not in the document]
forresttanaka: It’s closer, but it’s still not prettier compliant. I still see several changes when I save the file.

forresttanaka (changes requested):
I really love the new wrapper component on the pages. That makes maintaining those pages a lot easier. Getting a lot closer to what we need. We still need prettier compliance, and I’d like to see a justification for the Fetch class instead of adding functionality to the Request class.

components/edit.js
dougli1sqrd: it seems kinda nice as an internal separation in the file? But I can copy it back into the EditableItem directly if you'd prefer.

forresttanaka: Is this a different case from other components that don’t use their own custom hook for their implementation?

dougli1sqrd:
Yeah I definitely don't have a good argument for separating the Fetch class from the Request class. I only separated it because they have different header/auth constructions, and didn't quite know what the best approach would be to combine the two types of auth (cookie vs session). Maybe we can make that a separate issue, combining the two kinds of auth into one place?

forresttanaka:
> Maybe we can make that a separate issue, combining the two kinds of auth into one place?

It’ll take some thought to do this right, so I have no problems with that proposal.

components/edit.js
dougli1sqrd: The intention here was that you could use any "safe" HTTP methods here (GET, HEAD, OPTIONS, and TRACE): from https://restfulapi.net/http-methods/.

But I guess there's strictly speaking nothing stopping you from saying "PUT".

If I was using a typed language, I'd maybe do something like:

```rust
enum HttpMethod {
    GET,
    HEAD,
    OPTIONS,
    TRACE,
    PUT,
    POST,
    PATCH
}

enum SafeMethod {
    GET,
    HEAD,
    OPTIONS,
    TRACE
}

impl From<SafeMethod> for HttpMethod {
    fn from(safe: SafeMethod) -> HttpMethod {
        match safe {
            SafeMethod::SAFE => HttpMethod::SAFE,
            SafeMethod::HEAD => HttpMethod::HEAD,
            SafeMethod::OPTIONS => HttpMethod::OPTIONS,
            SafeMethod::TRACE => HttpMethod::TRACE
        }
    }
}

fn getObject(path: Path, method: SafeMethod) -> Response {
    let http_method = HttpMethod::from(method);
    let headers = Headers::make_headers(). // ...etc
    fetch(path, http_method, headers)
}
```
In this way you'd guarantee that you could *only* use a safe http method with getObject. You type check on SafeMethod, which gets converted to a regular HttpMethod which gets sent to fetch. (I mean to be clear I'm making up this api in rust, just basing it off the javascript). But in JS you can just pass in anything, so I'm not the most aware what the best way to enforce something like this.

Does typescript do things like this?

forresttanaka: I believe Typescript lets you do that kind of type checking at build time. I decided against Typescript since we’re all new to NextJS and Tailwind CSS already, and didn’t want to add yet another thing that’s still somewhat controversial in the Javascript community.

components/edit.js
dougli1sqrd: what about `setSaveErrors`? Just trying to indicate that the errors are from a "save".

forresttanaka: I think it’d be less confusing for future readers to just comment that the errors come from a save, since this is so unusual for state naming. I have a personal policy of commenting 100% of my states, instead of just those that I deem unclear at that moment, when they’ll probably _all_ be unclear in a few months or less.

forresttanaka: Also, I learned from Keenan not to be _too_ shy about long-ass explanatory variable names. I try to avoid it just because it’s kind of a pain, but Keenan’s Javascript is very clear partly for that reason.

components/edit.js
forresttanaka: Maybe at least make the default value for the method argument "GET" since the other possible values are quite rare, at least for us so far.

components/edit.js
forresttanaka: Another way to do this is to pass a function to `useState()` so that the callback calls a function that only gets executed the first render. But for your case, I think what you have here is better. But for other cases where a function initializes a state, you can do:
```
const [state, setState] = useState(() => f())
```
So in this example, `f()` only gets called on the first render, and its return value initializes `state`.

forresttanaka (approved):
OK, this is cool — let’s see how it does in QA! I’m really excited to see this go in. I have issues with the UI still, but Idan said to take care of it in another ticket.

PR #42, opened 2022-06-08
IGVF-93 Add authentication to Cypress tests on CircleCI

forresttanaka:
* I added the timeout change in .eslintrc.json within Cypress integration because the access-keys test takes far longer than the default four seconds to render the access-keys page while running Cypress in CircleCI.
* I didn’t include any server-side environment variable use. I think you wanted to use those for directing the front end to use different back ends, but I’d like to do that in a separate ticket.
* With Cypress 10.0.3, CircleCI would hang infinitely after the first Cypress test. This didn’t happen with Cypress 10.0.3 locally. I reverted to Cypress 9.7.0 and it doesn’t hang anywhere.
* I updated non-Cypress npm packages to their current latest versions.


keenangraham (approved):
LGTM.

PR #43, opened 2022-06-08
IGVF-172 Add pages for ontology terms and link to them from sample pages

forresttanaka:
* I added onto this ticket the animation of the navigation groups, so that hierarchical menus animate when they drop down. I share the code with the code that handles the mobile navigation.
* I copied and adapted the encoded code to handle the ontology links.


components/ontology.js
```diff
@@ -0,0 +1,45 @@
...
+  CL: "http://www.ontobee.org/ontology/CL?iri=http://purl.obolibrary.org/obo/",
+  CLO: "http://www.ontobee.org/ontology/CLO?iri=http://purl.obolibrary.org/obo/",
+}
+
+/**
+ * Display the ontology term ID given in `termId`, and link to a corresponding site if the prefix
+ * of the term ID has one. Any term IDs with prefixes not matching any in the `urlMap` global
+ * variable simply display without a link. All term IDs are of the form XXX:nnnnnnn...
+ */
+export const OntologyTermId = ({ termId }) => {
```
dougli1sqrd, line 20: It looks like the only place we get an ontology term CURIE is from `{ontologyTerm.term_id}` above. How is the `ontologyTerm` defined? Especially since these are ontology terms and not instance IDs (like a database ID: `FB:12335`) and if we knew how these `term_id`s are assembled we could nearly guarantee they have only one colon in the ID. Additionally, one strategy when dealing with CURIEs I've used before is to parse them into the "namespace" and "local ID" part, so you just have an object representing a correct ID. Maybe this compound object could be recieved by OntologyTermId instead, simplifying the checks?

dougli1sqrd (approved):

components/ontology.js
forresttanaka: @dougli1sqrd
> parse them into the "namespace" and "local ID" part, so you just have an object representing a correct ID
You mean anywhere we deal with an ontology term, pass around an object like:
```
{
  namespace: "EFO",
  id: "123"
}
```
?

forresttanaka: …or even a class if that’s not overkill.

dougli1sqrd: Yeah something like that!

forresttanaka:
* I really like your suggestion on using an object to track prefix and value, and ran with it to use a class to manage CURIEs, and extended it specifically for ontology terms. I could see an argument that this is overkill, but I do think it works well, and should extend to dbxrefs in a ticket that I have partially done now. dbxrefs have the same concerns but much more complicated URL composition than these ontology terms do.
* I also have other changes because igvfd renamed disease-ontology-terms to phenotype-ontology-terms, so that caused many Cypress tests to fail. I updated this branch to make them pass. Note that the sample-derived types (e.g. cell-lines) still refers to them as diseases even though they link to phenotype-ontology-terms now.

dougli1sqrd (dismissed):

forresttanaka:
@dougli1sqrd Since Jennifer found viable links for the DOID, HP, and MONDO prefixes, I added URLs for that. I also updated the now-non-functioning URLs for CL and CLO. Wouldn’t surprise me if we had to update those again later, but they seem to work for now. That’s all the changes, and I rebased to the tip of dev.

dougli1sqrd (dismissed):

libs/curie.js
```diff
@@ -2,34 +2,52 @@
  * Abstract base for mapping for CURIE-style identifiers (e.g. EFO:0002067) to links.
  */
 export default class Curie {
+  #curie
+  #prefix
+  #id
+
   constructor(curie) {
```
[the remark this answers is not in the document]
forresttanaka: I think I’ll leave this as is since any invalid curies would be very rare, as the schemas have strict regexes controlling these. The only realistic error situation would involve a schema with a curie the front end didn’t know about, in which case we’d still be able to display the original string, just without a link.

I need to rebase this, so this will probably show up for code review again.

dougli1sqrd (approved):

forresttanaka:
@dougli1sqrd igvfd updated the name of a property, and this branch used the old name. I modified this branch to now use the new names.

dougli1sqrd (approved):

dougli1sqrd (dismissed):

PR #45, opened 2022-06-21
IGVF-197 Stop redirecting when reloading a page

forresttanaka:
I removed the `<Session>` redirect. I don’t know why I thought I needed that before. For the most part it caused no harm, but it also didn’t help anything. But if you reload a non-home page or directly enter a URL to a non-home page, it reloads to some other page, often (but not always) the home page. Now it just stays on the page you loaded or reloaded.

This also fixes the grammar error you found before because the code right there changed, so I had to change what it said anyway.

### DIFF
--- a/components/session-context.js
+++ b/components/session-context.js
@@ -80,7 +80,7 @@ export const Session = ({ children }) => {
             isServerAuthPending.current = false
           } else {
             // Auth0 and the server authenticated successfully. Set the signed-in session object in
-            // the session context to that any downstream component can retrieve it without doing a
+            // the session context so that any downstream component can retrieve it without doing a
             // request to /session.
             getSession().then((sessionResponse) => {
               setSession(sessionResponse)


components/session-context.js
```diff
@@ -86,24 +79,18 @@ export const Session = ({ children }) => {
             logout({ returnTo: `${BACKEND_URL}/auth-error` })
             isServerAuthPending.current = false
           } else {
-            // Auth0 and the server authenticated successfully. Get the signed-in session object.
+            // Auth0 and the server authenticated successfully. Set the signed-in session object in
+            // the session context to that any downstream component can retrieve it without doing a
```
emro, line 83: Should this be "so that any downstream"?

emro (approved):
Looks great!

components/session-context.js
forresttanaka: Oh yeah — dang. I’ll change that since we’re not urgent.

emro (approved):

PR #46, opened 2022-06-28
IGVF-211 Remove propTypes warning with unknown collection types

forresttanaka:
Remove the label requirement.


emro (approved):
Looks good

PR #51, opened 2022-07-12
IGVF-219 Display differentiation treatments

forresttanaka:
Adds a display of the new `differentation_treatments` objects in Differentiated Cell and Differentiated Tissue objects as a table of treatments, in addition to the existing Treatment table.


emro (commented):
Looks great!

emro (approved):

PR #52, opened 2022-07-14
IGVF-234-node-installation-instructions

ottojolanki:


cdk/README.md
```diff
@@ -1,5 +1,25 @@
 ## Infrastructure
 
+Install Node.js 16:
```
keenangraham, line 3: Can you collapse this down to one or two code blocks and use concise comments instead? Think it might group better with the other sections.

> Install Node.js 16 using `nvm` (Node Version Manager):
>```bash
> # Install nvm.
> $ curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.1/install.sh | bash
>```
> In new terminal:
> ```bash
> # Install and use node 16.
> $ nvm install 16
> $ nvm use 16
> # Check version.
> $ node -v
> ```

keenangraham (changes requested):

cdk/README.md
ottojolanki: yup

keenangraham (approved):

PR #56, opened 2022-07-29
move-pre-commit-yml

keenangraham:


forresttanaka (approved):
It’s working for me, including in the Git Tower GUI. If we have issues later, we can fix them then, I think.

PR #57, opened 2022-08-01
IGVF-252 Add gene object and collection pages

forresttanaka:
I think from now on we have to name the object files in /pages/ to [id].js because igvfd has been changing its paths to not use uuids. Things still work, but it seems misleading to have params.uuid contain something that’s not a uuid. We might even want to go back and rename the other object pages to [id].js, at least if they need modification for another ticket.

These gene pages follow the same pattern as existing object and collection pages, so code review shouldn’t take that much effort. I did add a new icon for navigation.

### DIFF
--- a/cdk/infrastructure/config.py
+++ b/cdk/infrastructure/config.py
@@ -11,6 +11,7 @@ config: Dict[str, Any] = {
     'environment': {
         'demo': {
             'pipeline': 'DemoDeploymentPipelineStack',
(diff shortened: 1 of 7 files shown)


dougli1sqrd (approved):

forresttanaka:
The merge conflict is only in package-lock.json, so I’ll have npm rewrite that before merging.

forresttanaka:
Oh, it’s also in app.spec.js, so I’ll fix that momentarily.

PR #58, opened 2022-08-03
IGVF-237 Update Cypress to version 10

forresttanaka:
Cypress changed both the directory structure as well as the spec file names under the /cypress directory. With these changes, it now works for:

* Local without Docker
* Local with Docker on Apple Si.
* CircleCI


docker/cypress/Dockerfile
```diff
@@ -4,7 +4,7 @@
 
 FROM --platform=linux/arm64 node:16.13.0-bullseye as builder
 
-ARG VERSION=9.4.1
+ARG VERSION=10.4.0
```
keenangraham, line 7: I think we can delete this whole file, and the README in this directory, since `docker-compose.cypress-m1.yml` relies on official image now, and this is no longer used? 

Looks like we could probably use same `yml` locally and on circle, except environment variables are handled slightly differently. Maybe easier to keep separate for now.

keenangraham (dismissed):
Nice!

docker/cypress/Dockerfile
forresttanaka: Yeah, I tried to unify the environment variable use between those and couldn’t find anything that worked.

forresttanaka: I removed those two files, and all tests still work fine.

PR #59, opened 2022-08-03
IGVF-244 Change h1 titles to a bolder style

forresttanaka:
igvf-ui currently uses a very fine large typeface for the page titles, both the collection pages (e.g. /awards) and individual pages (e.g. /awards/HG012012).

This ticket changes that to a bolder font. To avoid looking too overwhelming on the page, I also make the text slightly smaller, and slightly gray instead of black.


forresttanaka:
The branch conflict is only in package-lock.json, so I’ll just have npm rewrite that before merging.

mingjiecn (approved):
Looks good!

PR #64, opened 2022-08-15
IGVF-95 Document UI

forresttanaka:
While the number of changed files is large, by far the most important parts are the new files under `components`. The rest are mostly files to test the new components, and pages that use the new components that mostly all make exactly the same changes to display these new components.

### DIFF
--- a/components/attachment-thumbnail.js
+++ b/components/attachment-thumbnail.js
@@ -1,0 +1,584 @@
+/**
+ * Displays a clickable thumbnail for an attachment, which could either appear as an icon
+ * representing the type of the document, or a preview image for attachments types that browsers
+ * can display directly in the page. Clicking icon thumbnails brings up the attachment in a browser
+ * tab for attachment types that browsers can display in a tab, or downloads the attachments for
+ * other attachment types. Clicking image thumbnails brings up the full-size preview image as an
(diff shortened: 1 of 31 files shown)


components/attachment-thumbnail.js
```diff
@@ -0,0 +1,584 @@
...
+ * Display an image thumbnail for an attachment that's viewable in the browser. When the user
+ * clicks the thumbnail, display the attachment image preview as a modal overlay on the page.
+ */
+const ImageThumbnailAndPreview = ({ attachment, ownerPath, alt, size }) => {
+  // True if the attachment image preview is visible.
+  const [isPreviewOpen, setIsPreviewOpen] = useState(false);
+
+  return (
+    <>
+      <button
```
dougli1sqrd, line 474: How come we don't use our Button component in components?

dougli1sqrd (dismissed):

components/attachment-thumbnail.js
forresttanaka: Good question. We do use the Button component in components, like below for the close button. But this one is really unusual because it doesn’t have a color outline with rounded corners nor text or icon content. It’s supposed to look like the linked icon.

PR #65, opened 2022-08-22
IGVF-284 Change biosample display for schema update

forresttanaka:
Since `biosample.embryonic` is valid and displayable, I had to check for the `embryonic` property’s existence using `in`.


mingjiecn (approved):

forresttanaka:
@mingjiecn Wranglers felt the age display on biosample-derived objects could give people the wrong impression, so I added this explanatory text after the age display and a very small component to display them in a standard way anywhere that a bit of extra explanation could help.

PR #66, opened 2022-08-23
IGVF-269-clean-up-demos-automatically

keenangraham:


cdk/infrastructure/config.py
```diff
@@ -4,17 +4,24 @@
...
 from typing import Optional
+from typing import Tuple
 
 
 config: Dict[str, Any] = {
     'environment': {
         'demo': {
             'pipeline': 'DemoDeploymentPipelineStack',
+            'tags': [
+                ('time-to-live-hours', '72'),
```
hitz, line 17: is there anyway to pass this as a command-line arg with a default?

keenangraham: It's possible but since there are two deployments here, one local that deploys the pipeline, and one in the pipeline that deploys the application, you would have to keep track of the args passed in locally and also pass them into the deploy command used by the pipeline.

The most important reason to do this though is so arguments used to make a certain deployment are committed in code, and there's no confusion later about _exactly_ how something was deployed and configured, and its current state for debugging.

PR #71, opened 2022-09-07
IGVF-273 Add UI to display and edit pages of arbitrary text

forresttanaka:
* Pages can have any path, so pages/[path].js handles them. If `<FallbackObject>` detects through the @type that this represents a page, it invokes the `<Page>` component. This doesn’t get included under the `pages` directory because doing so would add a /page/ path to the NextJS router that we don’t want. So the `<Page>` component falls under the `components` directory.
* I removed the h-screen Tailwind CSS class as it caused difficult-to-understand issues with page layout.
* This branch has a temporary ability to add pages that we might remove once IGVF-261 goes in.
* Some changes come from an update to heroicons 2.0, which changed the name of some icons we use.
* authentication.js has a change to use state instead of a ref for keeping track of the stable authentication  state. I don’t know why I used a ref, and using it caused components using `useAuthenticated` to not get updated about the authentication state.


dougli1sqrd (commented):
Wow lots here! I don't think I had much to say about specific code things - there's just a lot here and it all looks really great by and large! 

And I'm not fully sure where the border is between code review and QA review is so feel free to take this with a grain of salt, but I noticed a few bits of awkward workflow in using and editing the pages.
* If you cancel creating a page it brings you back to `/add-page` with nothing on it. It probably should take you back to `/pages`
* To me, the arrows on the `Add block above` and below and the button positions aren't super intuitive. My gut wants the above button to somehow be literally be "above" - maybe above the block content? 
* In the page view which shows all the existing pages, I found myself trying to click on the names of the pages to get to them, but only the `>>` button works. But I guess that's all over IGVF.
* Are there plans to make a page show its children pages?

Right now all our components are in one directory, `components` and I see you're adding a directory page-plugin. What are your thoughts on breaking up the components directory into sub directories? Like we have at least two button component files now, should they eventually go in a `button` directory or something? Just curious what the JS React party line is and what you like to do!

forresttanaka:
> If you cancel creating a page it brings you back to /add-page with nothing on it. It probably should take you back to /pages

I’ll take a look at that this morning. If a change is straightforward, I’ll put it back to In Progress.

> That gives me an idea, but not for this ticket as the changes might take a while, and Khine is waiting for this ticket so she can complete one of her tickets.

>  found myself trying to click on the names of the pages to get to them, but only the >> button works

We might change the design, and I’ve already changed it once. Ben said in a meeting he was fine with how it works here. I could see changing it to the way encoded works, but we’ll see.

> Are there plans to make a page show its children pages?

You mean when a user is looking at a page, or when editing a page? If viewing a page, I’d guess most people wouldn’t want that — a page should just display the page. If you mean when editing, what would that let the editor do?

> What are your thoughts on breaking up the components directory into sub directories?

We’ll do that, like we did over time with encoded. Typically if one thing starts moving into multiple files, we’ve done that. In fact I’ll be doing that in IGVF-302 because the collection components have grown into four files there, and maybe more. So yeah, I think a whole bunch of things will move to subdirectories of /components/ over time.

dougli1sqrd (dismissed):

PR #73, opened 2022-09-12
IGVF-294-gene-id-UI

khzlin:


pages/genes/[id].js
```diff
@@ -110,3 +110,19 @@ export const getServerSideProps = async ({ params, req }) => {
...
+    </a>
+  );
+};
+
+EnsemblLink.propTypes = {
+  // GeneID to display as a link
+  geneid: PropTypes.string.isRequired,
+  // Metadata that affects certain dbxrefs
+  taxa: PropTypes.string.isRequired,
+};
```
Comment (forresttanaka) [inline pages/genes/[id].js:128]: This component looks good! Could you move it above the `Gene` component? `getServerSideProps` is server-only code (NextJS looks for that function name specifically and removes it from browser code) and I’d like to keep server-only code at the bottom of files.

forresttanaka (changes requested):
Looks nice! I don’t know if we have any ENTREZ dbxrefs in the alpha database or not, but no problem adding them for processing here. The only change is to move the location of your new component so that server-side code stays at the bottom.

pages/genes/[id].js
Comment (khzlin) [reply pages/genes/[id].js:128]: I'll move the component to the top. Also, we currently don't have any ENTREZ dbxrefs in alpha database but will have them soon after we figure out a new gene loading script (wrangling).

forresttanaka (approved):
Thanks!

forresttanaka:
BTW, the Cypress error will get fixed in a branch that’s still in code review. No need to try to get it to pass right now.

PR #82, opened 2022-10-03
Igvf 261 add button

dougli1sqrd:
* Add button to add an object on each schema item in the schema list view and on each schema page
* The schema button just links to `/<collection>/#!add`, generating the Add editor
* Each collection page has a `<AddableItem ...>` surrounding the main content which detects when we're trying to add an object to the collection, changing the page to be the json editor.
* Some code is repeated between edit.js and add.js, but I tried to reuse when it was simple. There are definitely some opportunities to refactor
* Especially in the schema list view the Add buttons look quite clunky
* in lib/profile.js there is a map that links the schema id to it's collection id. I tried using the fact that the server will redirect something like `/human_donor` to `/human-donors` (the actual collection url) but FetchRequest was only getting the 301 redirect response instead of fetch following the redirect and returning the contents at the correct url, so this was the fastest/easiest way I came up with in the meantime to get something working.
* I tried to get the Cancel button to go back instead of displaying the underlying object so that when we're at the schema page and we Add, cancelling would bring us back to the schema page. But this proved slightly more difficult because of previous design decisions with SaveCancelControl and decided to defer that until later.

### DIFF
--- a/components/add.js
+++ b/components/add.js
@@ -102,9 +102,7 @@ export const AddInstancePage = ({ collection }) => {
   /**
    * The text is the current editor text of the underlying Ace editor component.
    */
-  const [text, setText] = useState(() => {
-    JSON.stringify({}, null, 4);
-  });
+  const [text, setText] = useState(() => JSON.stringify({}, null, 4));
 
   const router = useRouter();
 
@@ -246,12 +244,11 @@ AddInstancePage.propTypes = {
 };
 
 export const AddableItem = ({ collection, children }) => {
-  // return useEditor(collection, children, "add");
   const editing = useEditor("add");
   return editing ? (
     <AddInstancePage collection={collection} />
   ) : (
-    <>{children};</>
+    <>{children}</>
   );
 };
 


forresttanaka:
Holding on code review until tests pass.

components/add.js
```diff
@@ -0,0 +1,331 @@
...
+            : [
+                {
+                  description:
+                    "Error saving new item, ensure the fields are filled out correctly",
+                  keys: "Generic Error",
+                },
+              ];
+          setSaveErrors(errors);
+        }
+      });
```
forresttanaka, line 265: We now have a `FlashMessage` component to display temporary messages on the page, like error messages. The bulleted list seems like an odd UI that should get fixed here and in IGVF-201.

Screen Shot 2022-10-04 at 14 57 39

components/edit.js
```diff
@@ -60,12 +60,13 @@ EditableItem.propTypes = {
...
 };
 
-const SaveCancelControl = ({
+export const SaveCancelControl = ({
   cancelClick,
   saveClick,
   itemPath,
   saveEnabled,
 }) => {
+  // const router = useRouter();
```
forresttanaka, line 69: Would prefer not having commented-out code.

lib/fetch-request.js
```diff
@@ -181,7 +179,7 @@ export default class FetchRequest {
    * Determine whether this class object is being used on the server or not.
    * @returns {boolean} True if this class object exists on server, false for client
    */
-  get #isServer() {
+  get isServer() {
```
forresttanaka, line 182: Why is this public now? You don’t use it outside of `FetchRequest`.

pages/human-donors/[uuid].js
```diff
@@ -51,6 +51,7 @@ const HumanDonor = ({
   lab = null,
   parents,
 }) => {
+  console.log(donor);
```
Comment (forresttanaka) [inline pages/human-donors/[uuid].js:54]: Delete.

pages/profiles/[profile].js
```diff
@@ -9,13 +9,16 @@ import { DataPanel, DataAreaTitle } from "../../components/data-area";
 import buildBreadcrumbs from "../../lib/breadcrumbs";
 import errorObjectToProps from "../../lib/errors";
 import FetchRequest from "../../lib/fetch-request";
+import { AddItemFromSchema } from "../../components/add";
 
 const Schema = ({ schema, changelog }) => {
+  console.log(schema);
```
Comment (forresttanaka) [inline pages/profiles/[profile].js:15]: Delete

forresttanaka (changes requested):

components/add.js
```diff
@@ -0,0 +1,331 @@
...
+
+/**
+ *
+ * Given the schema ID, fetch the corresponding collection
+ * name. Example: schema id /profiles/human_donor.json maps
+ * to "human-donors"
+ */
+export const collectionPath = (schema) => {
+  return PROFILE_COLLECTIONS[schema["$id"]];
+};
```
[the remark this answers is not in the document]
dougli1sqrd: oh hm okay. I'll try that, but I was getting weird results and errors when the server would respond with a 301, but maybe there was something else going on in that particular instance.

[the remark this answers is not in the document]
dougli1sqrd: that would def be a lot simpler than this map ha

components/add.js
```diff
@@ -0,0 +1,331 @@
...
+    if (!isEdit && edit) {
+      setEditing(false);
+      router.replace(router.asPath);
+    }
+  }, [edit, router, action]);
+
+  const addpage = <AddInstancePage collection={collection} />;
+
+  const componentToShow = edit ? addpage : <>{viewComponent}</>;
+  return componentToShow;
```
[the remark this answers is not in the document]
dougli1sqrd: oh okay, should I just put this directly in <AddInstancePage> and have a children prop essentially? It would show if not `edit`, otherwise show the editor?

components/add.js
dougli1sqrd: maybe we should just add this to IGVF-201?

components/add.js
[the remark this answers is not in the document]
forresttanaka: Since you share this between add and edit, it seems fine to use this as a custom hook. But I think I would have it return `editing` and `setEditing` and have add and edit decide what JSX to render in response, whether the component unmodified, or the add/edit page. You’ll have more duplication, but very little. And if you did have a lot of duplication as a result of this, I would refactor the duplicate JSX into another component that wasn’t a custom hook.

components/add.js
forresttanaka: Sure, I think that makes sense.

components/add.js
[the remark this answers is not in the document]
dougli1sqrd: https://igvf.atlassian.net/browse/IGVF-319

components/add.js
dougli1sqrd: updated in IGVF-201

components/add.js
```diff
@@ -0,0 +1,289 @@
...
+  collection: PropTypes.object.isRequired,
+};
+
+export const AddableItem = ({ collection, children }) => {
+  // return useEditor(collection, children, "add");
+  const editing = useEditor("add");
+  return editing ? (
+    <AddInstancePage collection={collection} />
+  ) : (
+    <>{children};</>
```
forresttanaka, line 254: The semicolon shows up at the bottom of every collection page.

components/add.js
```diff
@@ -295,7 +251,14 @@ AddInstancePage.propTypes = {
 };
 
 export const AddableItem = ({ collection, children }) => {
-  return useEditor(collection, children, "add");
+  // return useEditor(collection, children, "add");
```
forresttanaka, line 249: Can we remove this commented-out code?

forresttanaka (changes requested):
Looks a lot easier to maintain! Just a couple smaller things and I think this should be ready for QA.

forresttanaka:
@dougli1sqrd changed this branch name to make a new PR, so this ticket has already been reviewed and merged into dev.

PR #83, opened 2022-10-12
IGVF-309-table-cell-renders

forresttanaka:
I broke my own rule very badly in this branch. Many Jest tests now fail coveralls coverage because the amount of time it will take to fix all these at this stage would be prohibitive IMO. I would like to have IGVF-324 as a high priority, maybe before IGVF-318.

### DIFF
--- a/pages/genes/[id].js
+++ b/pages/genes/[id].js
@@ -79,7 +79,7 @@ const Gene = ({ gene }) => {
             )}
             {gene.locations?.length > 0 && (
               <>
-                <DataItemLabel>Gene GeneLocations</DataItemLabel>
+                <DataItemLabel>Gene Locations</DataItemLabel>
                 <DataItemValue>
                   <ChromosomeLocations locations={gene.locations} />
                 </DataItemValue>
--- a/pages/genes/[id].js
+++ b/pages/genes/[id].js
@@ -9,6 +9,7 @@ import {
   DataPanel,
 } from "../../components/data-area";
 import DbxrefList from "../../components/dbxref-list";
+import ChromosomeLocations from "../../components/chromosome-locations";
 import PagePreamble from "../../components/page-preamble";
 import Status from "../../components/status";
 import { EditableItem } from "../../components/edit";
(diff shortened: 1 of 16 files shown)


pages/genes/[id].js
```diff
@@ -78,18 +79,9 @@ const Gene = ({ gene }) => {
             )}
             {gene.locations?.length > 0 && (
               <>
-                <DataItemLabel>Gene Locations</DataItemLabel>
+                <DataItemLabel>Gene GeneLocations</DataItemLabel>
```
Comment (mingjiecn) [inline pages/genes/[id].js:82]: Why change "Gene Locations" to "Gene GeneLocations"?

Comment (forresttanaka) [reply pages/genes/[id].js:82]: That’s a good question. Must have been a stray paste error.

mingjiecn:
How is the row length determined? Some rows has more content in certain cells but the rows are narrower compare to other rows. For example: external resources column is very narrow.
image

forresttanaka:
Column widths are determined by the browser. We can adjust it, and Ben and I talked about it this morning as something to work on in the future. But right now every column has the width the browser’s algorithm wants it.

mingjiecn:
OK. Then I am good after "Gene Locations" issue is fixed.

mingjiecn (changes requested):
fix paste error for "Gene Locations"

mingjiecn (approved):

forresttanaka:
@mingjiecn The package-lock.json file simply needs rebuilding to resolve the conflict, which I’ll do before merging to dev.

forresttanaka:
## Note for second round of code review:

* I didn’t take the downloaded TSV file into account with this change, so many of the cell contents in the TSV file no longer (or never) made sense, or didn’t take advantage of the new mechanisms. This change fixes that by using a similar but much simpler mechanism to have custom renderers for cells in the TSV file, though since they only output text and not React components, I call them “text composers” here instead of “renderers.”
* The non-breaking spaces I put into date displays in the first PR caused more problems than it solved when it came to TSV generation. I removed all of that.

mingjiecn:
NIH Institutional Certification is shown differently in list and download:
image
image

mingjiecn (changes requested):
NIH Institutional Certification is shown differently in list and download.

forresttanaka:
That appears to be a bug in the Numbers app. If you type “NIH00012" into a cell, it appears as “NIC 12.00.”

In Excel, it appears correctly. I don’t think there’s anything I can do about Numbers:
image

mingjiecn:
You are right!

mingjiecn (dismissed):

PR #84, opened 2022-10-17
Igvf-261-add-button (renamed branch)

dougli1sqrd:
I had to create a new branch name in order to satisfy the demo gods (underscores not allowed in branch names). Since PRs don't know how to track changed branch names, here's a new PR with the same code. Here is the old PR: https://github.com/IGVF-DACC/igvf-ui/pull/82

### DIFF
--- a/components/add.js
+++ b/components/add.js
@@ -276,7 +276,7 @@ AddableItem.propTypes = {
 export const AddItemFromSchema = ({
   schema,
   label = "Add Instance",
-  type = "primary",
+  type = "primary-outline",
   size = "sm",
 }) => {
   const { session } = useContext(SessionContext);
--- a/components/add.js
+++ b/components/add.js
@@ -10,13 +10,10 @@ import { useEffect } from 'react';
 import { SaveCancelControl, SavedErrors } from "./edit";
 import { useRouter } from 'next/router';
 import { empty } from 'empty-schema';
+import { PROFILE_COLLECTIONS } from '../lib/profiles';
 
 export const collectionPath = (schema) => {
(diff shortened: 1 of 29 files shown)


components/add.js
```diff
@@ -0,0 +1,306 @@
...
+    );
+  }
+};
+
+AddLink.propTypes = {
+  collection: PropTypes.object.isRequired,
+  label: PropTypes.string,
+  type: PropTypes.string,
+  size: PropTypes.string,
+};
```
forresttanaka, line 107: It’s not required, but you can catch more errors is you select the allowed values with [PropTypes.oneOf](https://reactjs.org/docs/typechecking-with-proptypes.html#proptypes)

forresttanaka (dismissed):
Odd you have to “run prettier.” It should run itself every time you save your file.

forresttanaka (dismissed):

forresttanaka (changes requested):
OK, now _I’m_ going to complain about a lack of gaps. There’s no gap between the Add button and the panel on individual schema pages. If you sign in and look at the Pages page, you can see an Add Page button with an `mb-1` gap between it and the panel. Could you add that gap to the schema add button as well? It’s also a different color, but I don’t know if I should make the “Add Page” button primary-outline or if you should make the Add button success-outline. We haven’t decided on button colors and styles overall yet.

dougli1sqrd:
I had to fiddle a little, but I got it

forresttanaka (approved):
Thanks. We need to get back to work on the design document so that similar things on different pages appear the same.

PR #85, opened 2022-10-26
IGVF-336-crashfix

dougli1sqrd:
* Pulls in @forresttanaka mapprofile nextjs api
* The mapprofile api will take a schema name and return the associated collection
* This collection is then used for the Add button logic
* Removed the hard coded mapping between schema and collection

### DIFF
--- a/components/profile-map.js
+++ b/components/profile-map.js
@@ -6,6 +6,28 @@ const ProfileMapContext = createContext({
   profileMap: {},
 });
 
+/**
+ * This component is used in _App to generate the mappings between schemas
+ * and their corresponding collection in the backend.
+ *
+ * Schemas define what types of objects can go inside a collection, so
+ * there's a natural correspondence between schemas and collections.
+ * For example the schema for /profiles/human_donor/ has a corresponding
+ * collection /human-donors/. If a component needs to know what
+ * collection corresponds to a given schema, it can use the ProfileMapContext
+ * with:
+ *
+ * const { profileMap } = useContext(ProfileMapContext);
+ *
+ * Then the mapping can be accessed with `profileMap[schemaId]` where
+ * schemaId is something like "human_donor", like above.
+ *
+ * This works by going through the nextjs api in the pages/api/mapprofile.js
+ * which provides the actual mapping.
+ *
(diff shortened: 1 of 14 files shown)


components/profile-map.js
```diff
@@ -0,0 +1,43 @@
...
+        })
+        .forEach((p) => {
+          const profileName = p["$id"].match(/^\/profiles\/(.+).json$/)[1];
+          // Get the mapped collection for the profile schema
+          api
+            .getObject(`/api/mapprofile?profile=${profileName}`)
+            .then((collect) => {
+              setMapping((current) => {
+                return { ...current, [profileName]: collect };
+              });
```
forresttanaka, line 30: Might be a good place to use implicit return, but that’s completely optional.

pages/api/mapprofile.js
```diff
@@ -0,0 +1,29 @@
...
+            .json({ error: "Bad Request" });
+        }
+      });
+    } else {
+      res.status(HTTP_STATUS_CODE.BAD_REQUEST).json({ error: "Bad Request" });
+    }
+  } else {
+    res.status(HTTP_STATUS_CODE.METHOD_NOT_ALLOWED).end();
+  }
+};
```
forresttanaka, line 27: If you haven’t yet, make sure to write a ticket for that warning you get in the browser console.

forresttanaka (dismissed):
Just a couple optional changes.

pages/api/mapprofile.js
dougli1sqrd: https://igvf.atlassian.net/browse/IGVF-338

forresttanaka:
Sorry @dougli1sqrd; could you make one more change to this branch — not for a problem, but to make another part of the code simpler? In ProfileMapContext could you provide a _second_ property: `profiles`. In it, just put the result of retrieving /profiles. Then in your `ProfileMapContext.Provider` wrapper, provide that as the second property in the `value` object.

forresttanaka:
@dougli1sqrd Thinking about it more, it makes sense to _not_ make that change in this branch since we’ll need another ticket to simplify the other code module, so we could just make the change I described in that branch.

forresttanaka (dismissed):
Awesome!

PR #88, opened 2022-11-09
IGVF-327-500-error updated FallbackCollection to use the Pager pattern for collections

dougli1sqrd:
I just updated the FallbackCollection code to use the newer pager pattern that the other collections use. I mirrored it after the `human-donors` collection page.


forresttanaka (approved):
Very nice!

PR #92, opened 2022-12-02
IGVF-318 Get rid of the console /api/mapprofiles error

forresttanaka:
Turns out this error was caused by not returning the getObjects promise from the mapprofiles API route function. I also changed the mapprofiles endpoint to use a dynamic route, so instead of `/api/mapprofiles?query=<profile>` we use `/api/mapprofiles/<profile>` which seemed more straightforward.

I also stopped requesting the access_key collection because you have to be signed in for this to work, and we don’t create access keys using the editor UI anyway. Keenan said we have to keep the access_key collection privileged.

### DIFF
--- a/components/profile-map.js
+++ b/components/profile-map.js
@@ -7,6 +7,14 @@ const ProfileMapContext = createContext({
 });
 
 /**
+ * $id values of schemas we don't need to retrieve the empty collection data from. Each comment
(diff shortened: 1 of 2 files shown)


forresttanaka:
I screwed up the ticket name again. I’ll close this and open another PR.

PR #96, opened 2023-01-20
IGVF-391 Implement design document button styling

forresttanaka:
* This branch implements the design document so that now only three (really four) button types exist: “primary,” “secondary,” and “warning.” In addition, you can use “selected” for attached buttons along with “secondary.”
* While this branch affects a large number of files, the vast majority involve adapting to the new button UI, so the changes don’t amount to much in most files. Jest tests take up another large set of changes. The bulk of the important changes reside in button.js, now in the `form-elements` directory instead of directly in the `components` directory. Since the button colors have completely changed, styles/globals.css has many constants deleted and added, and tailwind.config.css has its button styles redone to use these constants.
* The `TextField` and `Select` components also got revamped to work with the new buttons better, as well as also getting moved to the `form-elements` directory. These might show a lot of apparent changes but that results from moving them out of the `components` directory into the `components/form-elements` directory.
* Moved existing button, select, and text-input components to a new `form-elements` directory. Future form elements like radio buttons and checkboxes will go here. Anyone needing a form element just needs to import them from `components/form-elements` because of the index.js there. Likely the format of the text-input and select elements will change as the design document moves forward. The button styles likely won’t change in the near future.
* Removed the tooltip component. Might return if we write/acquire a good one that adjusts for screen position, width, and accessibility better. For now, I just avoid the need for tooltips — easy enough to do with touch interfaces as inspiration, where you can display always-visible instructions, make the control titles more explicit, or add a modal with instructions (e.g. report pages).


emro (approved):
Looks great, I like the new implementation and styles a lot.

PR #100, opened 2023-01-30
IGVF-372-search-conversion

forresttanaka:
At a high level, I added` /pages/search.js` and `/pages/report.js` for the new list and report pages. I got rid of the `/components/collections` as well as each of the `/pages/{page}/index.js` because now all collections pages just display JSON for the collection. Some of the files that were in `/components/collections` have moved to the `/components/search` or `/components/reports` directories.

I copied the query-string.js library from encoded, but changed the `negative` parameters to `polarity` parameters that should make key!=value query-string elements more clear and easier to work with.

### New Components Directories
Within the components directory, I now have new `/search`, `/search-results`, and `/report` directories with these roles:
* `/search`: components for rendering `/search?type=Type` pages that I usually call “list” pages.
* `/report`: components and functions for rendering `/report?type=Type` pages.
* `/search-results`: components and functions for both `/search` and `/report` pages.

### List Renderers
The `/components/search/list-renderer` directory has files with components to render the list items for each `@type` in a list view. In some cases multiple `@types` share the same renderer, such as all the Biosample-derived types.

Any future list-item renderers need to be have their own file in the `list-renderers` directory, imported into `/components/list-renderer/index.js` and then added to the `renderers` object. Once you do this, they appear in relevant list pages.

### Accessory Data
Some of the list-item renderers have something called “accessory data.” I use this for list-item renderers that display data from linked objects, but that igvfd does not embed in search results. `/pages/search.js` requests this data from igvfd on the NextJS server so that they have loaded by the time the browser sees any data — the browser code never requests accessory data after page load.

For any list-item renderer that needs accessory data, it has a `getAccessoryDataPaths()` method added to the renderer component that returns an array of paths to objects it needs to render. That array can contain paths to a variety of objects. For example, the Biosample list renderer returns paths to both biosample-term objects and lab objects in the same array. `/pages/search.js` then loads this data from `getServerSideProps()` and passes it in the JSON for the page.

If igvfd changes to have embedded properties in search results for certain object types, we might need to remove some paths from the corresponding list-item renderer’s `getAccessoryDataPaths()` method.

The report view doesn’t use nor load accessory data.

Unrelated to this, I got rid of the `<App>` cookie state as it has no use whatsoever. I also removed the collection view state since we no longer display collections.


components/no-collection-data.js
```diff
@@ -7,14 +8,21 @@ import GlobalContext from "./global-context";
...
  */
-const NoCollectionData = () => {
+const NoCollectionData = ({ pageTitle = "" }) => {
   const { page } = useContext(GlobalContext);
 
   return (
     <DataPanel className="my-0.5">
-      <div className="text-center italic">No {page.title} to display</div>
+      <div className="text-center italic">
+        No {pageTitle || page.title} to display
```
mingjiecn, line 17: This code doesn't seem to work properly.  The title is always "list items"

forresttanaka: You mean this, right?
image
That’s what I intended to display since search-list displays can have a mix of items. Do you think it should display something else?

mingjiecn: Oh, I thought you will show the name of search item. In that case, it should be OK.

pages/report.js
```diff
@@ -0,0 +1,253 @@
...
+  };
+
+  if (profiles) {
+    const pageTitle = composeSearchResultsPageTitle(searchResults, profiles);
+    const items = searchResults["@graph"];
+    const columns = generateColumns(searchResults, profiles);
+
+    return (
+      <>
+        {pageTitle && <PagePreamble pageTitle={pageTitle} />}
```
mingjiecn, line 190: should display Breadcrumbs on the top to make it consistent.

forresttanaka: Yeah, that’s true. I’ve added it to a list for a future ticket.

mingjiecn: Sounds good

mingjiecn (approved):

PR #105, opened 2023-02-08
IGVF-399 Add Google Analytics version 4

mingjiecn:


forresttanaka (changes requested):
This will seem really picky, but to me it seems odd (to me at least) to have a `<Script>` component before the `<Head>` component, even though it completely works. I just expect the `<Head>` component first, since the `<head>` tag goes first in the generated HTML. In the [NextJS documentation](https://nextjs.org/docs/messages/no-script-component-in-head), they put the `<Script>` component after the `<Head>` component. I think it produces exactly the same HTML as what you have.

mingjiecn:
> This will seem really picky, but to me it seems odd (to me at least) to have a `<Script>` component before the `<Head>` component, even though it completely works. I just expect the `<Head>` component first, since the `<head>` tag goes first in the generated HTML. In the [NextJS documentation](https://nextjs.org/docs/messages/no-script-component-in-head), they put the `<Script>` component after the `<Head>` component. I think it produces exactly the same HTML as what you have.

I will move it and test it.

forresttanaka (approved):
Thanks!

PR #106, opened 2023-02-10
IGVF-430-check-config-on-pr

ottojolanki:


cdk/tests/unit/test_config.py
```diff
@@ -1,6 +1,15 @@
 import pytest
 
 
+def test_demo_environment_not_changed_when_opening_PR():
```
keenangraham, line 4: Should be lower case.

.circleci/config.yml
```diff
@@ -115,4 +115,4 @@ jobs:
           path: ~/igvf-ui
       - run: pyenv global 3.9.7
       - run: pip install -r requirements.txt -r requirements-dev.txt
-      - run: pytest tests/
+      - run: pytest tests/ -k "not test_demo_environment_not_changed_when_opening_PR"
```
keenangraham, line 118: Can we put this test somewhere else that's only run when opening PR instead of mixed in with the other tests? I don't want to have to add this special condition when running tests locally when the demo config is changes etc.

Maybe just make it a one-off Python script that loads the config and raises an error instead of running it as a test? I think that would cause the PR check to fail and would make it easier to run selectively.

keenangraham (changes requested):

.circleci/config.yml
keenangraham: Add it to this folder for example if you would like: https://github.com/IGVF-DACC/igvf-ui/tree/dev/cdk/commands

keenangraham (approved):
Nice.

keenangraham (approved):

PR #107, opened 2023-02-13
IGVF-381 Add in vitro system object, list, and report pages

forresttanaka:
This mostly copies other biosample-derived pages such as Primary Cells, so the displays should appear basically the same.


mingjiecn (approved):

mingjiecn:
LGTM

PR #111, opened 2023-02-16
IGVF-403-file-set

mingjiecn:


forresttanaka (changes requested):
It looks really good! Hopefully the list-item mechanism is easy enough to understand. I have a few things to change or suggestions. It looks like you have a couple lines of Jest coverage missing which shouldn’t be too hard to fix. It still passes because coveralls has a minimum threshold, but good to keep it at 100% if we can. Also I tried to configure the ESLint config to not allow snake-case variables while allowing snake-case properties from igvfd, but I can’t seem to get that configured right yet.

forresttanaka (dismissed):
Nice! Files should be fairly similar I’d guess.

mingjiecn:
> Nice! Files should be fairly similar I’d guess.

Let's hope so!

forresttanaka (dismissed):
I had to look up how .flat() works. That’s really cool! Thanks for the Javascript lesson. Code looks really good.

forresttanaka:
You’ll have to rebase to dev too. I squeezed in a couple commits today.

forresttanaka:
@mingjiecn Rebase this branch and push, and just do a very quick sanity test before setting to Merge Ready.

forresttanaka (approved):
I’ll just approve and merge. The changes are pretty small and logical, and a quick test shows it works properly.

PR #112, opened 2023-02-17
IGVF-404-file-collection

mingjiecn:


components/search/list-renderer/index.js
```diff
@@ -37,6 +37,7 @@ import RodentDonor from "./rodent-donor";
 import TechnicalSample from "./technical-sample";
 import Treatment from "./treatment";
 import User from "./user";
+import File from "./file";
```
forresttanaka, line 40: Not something you need to change, but I try to keep imports sorted by file name so it’s easy to find at a glance, for future PRs. That’s why imports for most files are broken up into sections (// node_modules, // components, etc.) so sorting by file name is easier (sections don’t apply to this file), and it’s easier to find something just by looking.

components/search/list-renderer/file.js
```diff
@@ -0,0 +1,49 @@
...
+File.propTypes = {
+  // Single file search-result object to display on a search-result list page
+  item: PropTypes.object.isRequired,
+  // Accessory data to display for all search-result objects
+  accessoryData: PropTypes.object,
+};
+
+File.getAccessoryDataPaths = (files) => {
+  const labs = files.map((file) => file.lab);
+  return labs;
```
forresttanaka, line 48: You can also just `return files.map…` just to save some typing. Otherwise doesn’t make a difference.

forresttanaka (approved):
This is great! Using the common file data items is a really nice touch. I have a couple comments, none of which you need to do anything about.

PR #118, opened 2023-03-01
IGVF-455-source

mingjiecn:


forresttanaka (approved):
😎
Would the title work better as the title of Source object pages instead of the name? The title seems more readable, and that would more closely reflect the list results.

mingjiecn:
> 😎 Would the title work better as the title of Source object pages instead of the name? The title seems more readable, and that would more closely reflect the list results.

Agree. Will update. Thank you!

PR #119, opened 2023-03-06
IGVF-322-edit-tests

dougli1sqrd:
This adds tests for the add and edit button.

There are jest tests for the json editor used in both the add and edit button/pages.

Additionally there are cypress tests for the add and edit buttons.

As a note, the edit and add components have line coverage disabled as jest does not cover the full range since the ace editor component is difficult to unit tests. Thus the cypress tests help supplement the unit tests.


.eslintrc.json
```diff
@@ -21,6 +22,10 @@
         "named": "never",
         "asyncArrow": "always"
       }
-    ]
+    ],
+    "testing-library/await-async-query": "error",
+		"testing-library/no-await-sync-query": "error",
+		"testing-library/no-debugging-utils": "warn",
+		"testing-library/no-dom-import": "off"
   }
```
forresttanaka, line 30: Could you change the indents of the previous three lines from tabs to spaces? The prettier plugin for VSCode should automatically make these spaces.

forresttanaka (changes requested):
Before I code review, could you get the CircleCI tests to pass? It looks like quite a few holes in the Jest coverage too.

Make sure your VSCode is configured for prettier. Right now if I save most any file in your branch, I get a whole bunch of modifications to the file because of prettier violations. That’ll cause problems for the next person who modifies those files.

forresttanaka:
Also, could you resolve the conflict in package.json? You can regenerate package-lock.json from scratch to resolve that one.

dougli1sqrd:
Yeah will do

cypress/e2e/edit.cy.js
```diff
@@ -0,0 +1,162 @@
...
+      .type('"/labs/christina-leslie/"\n')
+      .type("{backspace}{rightArrow},\n")
+      .type('"viewing_groups": [\n')
+      .type('"IGVF"\n')
+      .type("{backspace}{rightArrow}\n")
+      .type("{backspace}{backspace}");
+    cy.contains(`${now}`).should("exist");
+
+    cy.get("button").contains("Save").click();
+    cy.reload();
```
forresttanaka, line 106: The following h1 text fails locally for me because it reloads the page before elasticsearch reindexes the new title. Could you put a `cy.wait(1000)` before the `cy.reload()`? That should make this test more reliable.

cypress/e2e/access-keys.cy.js
```diff
@@ -7,6 +7,8 @@ describe("Exercise access keys", () => {
       defaultCommandTimeout: 30000,
     },
     () => {
+      cy.log("***************");
+      cy.log(`HELLLOOO auth url ${Cypress.env("CYPRESS_AUTH_URL")}`);
       cy.log(`CYPRESSENV: ${Cypress.env("AUTH_USERNAME")}`);
```
forresttanaka, line 12: You can delete the cy.logs. You can even delete the “CYPRESSENV” one that I added. Should really get rid of all those. I’d forgotten about them.

cypress/e2e/add.cy.js
```diff
@@ -0,0 +1,35 @@
...
+      .type('"description": "Very Good Lab at Very Good University",\n')
+      .type(`"name": "lab-${t}",\n`)
+      // This is the cypress user
+      .type('"pi": "/users/4a63f82a-2e25-4790-bfcc-d183f8ca8869/",\n')
+      .type('"institute_label": "Very Good University"');
+    cy.contains("http://verygood.university.edu").should("exist");
+    cy.get("button").contains("Save").click();
+    cy.get("h1")
+      .contains("Cypress Testing, Very Good University")
+      .should("exist");
```
forresttanaka, line 33: You can end this clause with the `contains()` and delete the `should()`. `contains()` is designed to fail with useful information if that string doesn’t exist.

lib/general.js
```diff
@@ -32,3 +32,20 @@ export function isValidUrl(url) {
 export function urlWithoutParams(url) {
   return url.split("?")[0];
 }
+
+export function sortedJson(obj) {
```
forresttanaka, line 36: It’d be nice to document this function’s purpose.

package.json
```diff
@@ -7,13 +7,15 @@
...
-    "cypress:run": "cypress run --record --key 083e76f3-5f44-43f0-b248-ace10abf0877"
+    "cypress:run": "cypress run --record --key 083e76f3-5f44-43f0-b248-ace10abf0877",
+    "format": "prettier --write ."
   },
   "dependencies": {
     "@auth0/auth0-react": "^1.10.1",
     "@headlessui/react": "^1.6.4",
     "@heroicons/react": "^2.0.10",
     "ace-builds": "^1.5.0",
+    "babel-plugin-transform-dynamic-import": "^2.1.0",
```
forresttanaka, line 18: What’s this for?

forresttanaka (changes requested):
Looks pretty good. Still lots of holes in the Jest test that didn’t exist before, but we have to move on, so let’s just gradually fix those in other tickets, even unrelated ones, until we’re back to 100% on everything.

cypress/e2e/edit.cy.js
```diff
@@ -0,0 +1,162 @@
...
+    cy.get("h1").then(($el) => {
```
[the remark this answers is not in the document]
dougli1sqrd: It was having a lot of trouble matching the contents in the h1 directly like that, but comparing the strings worked great. I'm not sure why it seemed to make a difference. I can remove the log statement though.

cypress/e2e/access-keys.cy.js
dougli1sqrd: I can also remove the AUTH_USERNAME log then

package.json
forresttanaka: I guess it’s for something, but I’m not sure why old tests that didn’t need it before now fail.

forresttanaka (changes requested):
Make just _one_ more change. You have to pin headlessui at 1.7.11, so remove the `^` in the semver in package.json, so it’s just "1.17.11" instead of "^1.17.11" and then rebuild package-lock.json from scratch and commit both of those. The one with the caret can still install later versions of headlessui.

forresttanaka:
@dougli1sqrd Could you rebase with dev (merged a couple other tickets this morning), do a quick test, and then set back to merge ready when you’re done?

dougli1sqrd:
updated

PR #120, opened 2023-03-06
IGVF-216-publication

mingjiecn:


forresttanaka (approved):
Very nicely done. I think if we someday get another place where we need to display a citation, we might want to put some of this code into its own component to do that, so we can just plonk that in wherever it needs to go without code duplication.

forresttanaka:
The icon for publications is the same as the one for editing an object. I think that’s OK, but something to note. I had been thinking about asking Eric to change the button to edit an object to the word “Edit,” but the “Add” button is the plus sign, which I think is the right way to do that.

PR #122, opened 2023-03-07
IGVF-453-biomarker

mingjiecn:


pages/biomarkers/[id].js
```diff
@@ -0,0 +1,115 @@
...
+  return (
+    <>
+      <Breadcrumbs />
+      <EditableItem item={biomarker}>
+        <PagePreamble />
+        <DataPanel>
+          <DataArea>
+            <DataItemLabel>Name</DataItemLabel>
+            <DataItemValue>{biomarker.name}</DataItemValue>
+            <DataItemLabel>quantification</DataItemLabel>
```
Comment (forresttanaka) [inline pages/biomarkers/[id].js:36]: The `q` is capitalized in this property’s schema title, so it should be capitalized here.

pages/biomarkers/[id].js
```diff
@@ -0,0 +1,115 @@
...
+      "name",
+      req.headers.cookie
+    );
+    return {
+      props: {
+        biomarker,
+        award,
+        lab,
+        gene,
+        pageContext: { title: biomarker.name },
```
Comment (forresttanaka) [inline pages/biomarkers/[id].js:109]: It’d be nice to make the page title the same as the title in the search results. If you do this, you could make a new biomarker.js file in the `lib` directory to share the function that composes the title between the list renderer and this file. It would also be nice to have this title in the breadcrumb for the object page, but we can’t do that until IGVF-274 (which maybe I should work on sooner than later).

forresttanaka (changes requested):
One very small required changes and a couple optional changes. You also need to resolve the new merge conflicts.

components/search/list-renderer/biomarker.js
```diff
@@ -0,0 +1,47 @@
...
+    Boolean
+  );
+  const title = titleElements.join(" ");
+
+  return (
+    <SearchListItemContent>
+      <SearchListItemMain>
+        <SearchListItemUniqueId>
+          <SearchListItemType item={biomarker} />
+          {title}
```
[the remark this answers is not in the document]
mingjiecn: I will use uuid then

lib/biomarker.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * Use the same title for biomarker search page and biomarker object page for that biomarker
+ */
```
forresttanaka, line 3: We’ve (“I’ve”?) been using JSDoc for normal function comments (as opposed to React component comments which have props documented in propTypes). That way when someone else using VSCode hovers over the function name where it gets called, they can see the parameters and return values documented in the popover without having to go to the function definition itself.

pages/biomarkers/[id].js
```diff
@@ -0,0 +1,117 @@
...
+      req.headers.cookie
+    );
+    const title = getTitle(biomarker);
+    return {
+      props: {
+        biomarker,
+        award,
+        lab,
+        gene,
+        pageContext: { title: title },
```
Comment (forresttanaka) [inline pages/biomarkers/[id].js:111]: As a shortcut, since the key and value variable names are the same, you can use `{ title }`. It results in the exact same operation as what you have but with less typing.

lib/biomarker.js
```diff
@@ -0,0 +1,10 @@
...
+
+export function getTitle(biomarker) {
+  const titleElements = [biomarker.name, biomarker.quantification].filter(
+    Boolean
+  );
```
forresttanaka, line 8: It looks like both `name` and `quantification` are both required in the schema, so you technically don’t need the Boolean filter. I could see an argument to keep the filter in case either of them get changed to non-required, though if they both become non-required, we’d need a fallback in case neither exists.

forresttanaka (approved):
All my comments are suggestions, though I do encourage but don’t require the `getTitle` name change and the JSDoc comment. Nice job with all these tickets keeping the Jest coverage at 100%. That will unfortunately change to a lot less than 100% probably some time this week because of IGVF-322, but we hope to get that back to 100% in the coming weeks through other tickets.

PR #124, opened 2023-03-10
IGVF-435-reconcile-human-donor

dougli1sqrd:
preliminary adding phenotypic features, donor references, and url to human donor object pages. Start of adding human donor phenotypic features and collections to search result items

### DIFF
--- a/components/common-data-items.js
+++ b/components/common-data-items.js
@@ -33,9 +33,7 @@ export function DonorDataItems({ donor, parents, children }) {
       {donor.ethnicities?.length > 0 && (
         <>
           <DataItemLabel>Ethnicities</DataItemLabel>
-          <DataItemValue>
-            {donor.ethnicities.join(", ")}
-          </DataItemValue>
+          <DataItemValue>{donor.ethnicities.join(", ")}</DataItemValue>
         </>
       )}
       {donor.sex && (
@@ -94,7 +92,9 @@ export function DonorDataItems({ donor, parents, children }) {
       {donor.url && (
         <>
           <DataItemLabel>URL</DataItemLabel>
-          <DataItemValue><Link href={donor.url}>{donor.url}</Link></DataItemValue>
+          <DataItemValue>
+            <Link href={donor.url}>{donor.url}</Link>
+          </DataItemValue>
         </>
       )}
     </>
--- a/components/common-data-items.js
+++ b/components/common-data-items.js
@@ -24,16 +24,24 @@ import { formatDate } from "../lib/dates";
 export function DonorDataItems({ donor, parents, children }) {
   return (
     <>
+      {donor.taxa && (
+        <>
+          <DataItemLabel>Taxa</DataItemLabel>
+          <DataItemValue>{donor.taxa}</DataItemValue>
+        </>
+      )}
+      {donor.ethnicities?.length > 0 && (
+        <>
+          <DataItemLabel>Ethnicities</DataItemLabel>
+          <DataItemValue>
+            {donor.ethnicities.join(", ")}
+          </DataItemValue>
+        </>
+      )}
       {donor.sex && (
         <>
           <DataItemLabel>Sex</DataItemLabel>
           <DataItemValue>{donor.sex}</DataItemValue>
-        </>
-      )}
-      {donor.taxon_id && (
-        <>
-          <DataItemLabel>Taxa Identifier</DataItemLabel>
-          <DataItemValue>{donor.taxon_id}</DataItemValue>
         </>
       )}
       {parents.length > 0 && (
@@ -73,6 +81,20 @@ export function DonorDataItems({ donor, parents, children }) {
           <DataItemValue>
             <AliasList aliases={donor.aliases} />
           </DataItemValue>
+        </>
+      )}
+      {donor.references?.length > 0 && (
+        <>
+          <DataItemLabel>References</DataItemLabel>
+          <DataItemValue>
+            <DbxrefList dbxrefs={donor.references} />
+          </DataItemValue>
+        </>
+      )}
+      {donor.url && (
+        <>
+          <DataItemLabel>URL</DataItemLabel>
+          <DataItemValue><Link href={donor.url}>{donor.url}</Link></DataItemValue>
         </>
       )}
     </>
--- a/components/phenotypic-feature-table.js
+++ b/components/phenotypic-feature-table.js
@@ -32,7 +32,11 @@ const phenotypicFeaturesColumns = [
     id: "feature",
     title: "Feature",
     display: ({ source }) => {
-      return <Link href={source.feature}>{source.feature.split("/").slice("-2")}</Link>;
+      return (
+        <Link href={source.feature}>
+          {source.feature.split("/").slice("-2")}
+        </Link>
+      );
     },
   },
   {
--- a/components/phenotypic-feature-table.js
+++ b/components/phenotypic-feature-table.js
@@ -1,0 +1,67 @@
+// node_modules
+import Link from "next/link";
+import PropTypes from "prop-types";
+// components
+import { DataGridContainer } from "./data-grid";
+import SortableGrid from "./sortable-grid";
+
+/*
+phenotypic feature
+lab	"/labs/j-michael-cherry/"
+award	"/awards/HG012012/"
+notes	"Phenotypic feature of body weight"
+status	"released"
+feature	"/phenotype-terms/NCIT_C92648/"
+quantity	58
+quantity_units	"kilogram"
+schema_version	"1"
+observation_date	"2022-11-15"
+creation_timestamp	"2023-03-06T23:31:20.467925+00:00"
+@id: /phenotypic-features/ae1b4a0b-78e6-af0a-8e6d-c0c9b45905fa/
+*/
+
+const phenotypicFeaturesColumns = [
+  {
+    id: "phenotypic_feature_id",
+    title: "Phenotypic Feature",
+    display: ({ source }) => {
+      return <Link href={source["@id"]}>{source.notes}</Link>;
+    },
+  },
+  {
+    id: "feature",
+    title: "Feature",
+    display: ({ source }) => {
+      return <Link href={source.feature}>{source.feature.split("/").slice("-2")}</Link>;
+    },
+  },
+  {
+    id: "quantity",
+    title: "Quantity",
+    display: ({ source }) => `${source.quantity} ${source.quantity_units}`,
+  },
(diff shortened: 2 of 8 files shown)


components/search/list-renderer/human-donor.js
```diff
@@ -31,6 +36,8 @@ export default function HumanDonor({ item: humanDonor, accessoryData }) {
         {lab && (
           <SearchListItemMeta>
             <div key="lab">{lab.title}</div>
+            <div key="collections">{collections}</div>
+            <div key="phynotypic-features">{phenotypicFeatures}</div>
```
forresttanaka, line 40: This won’t display collectinos or phenotypic features if the lab couldn’t be loaded, so you’ll probably want to put all three of these things in a non-sparse array and then display them in a loop inside `<SearchListItemMeta>`.

components/phenotypic-feature-table.js
```diff
@@ -0,0 +1,71 @@
...
+creation_timestamp	"2023-03-06T23:31:20.467925+00:00"
+@id: /phenotypic-features/ae1b4a0b-78e6-af0a-8e6d-c0c9b45905fa/
+*/
+
+const phenotypicFeaturesColumns = [
+  {
+    id: "phenotypic_feature_id",
+    title: "Phenotypic Feature",
+    display: ({ source }) => {
+      return <Link href={source["@id"]}>{source.notes}</Link>;
```
forresttanaka, line 28: Since `notes` isn’t a required property, should have a fallback to something that we know exists, like maybe uuid.

components/phenotypic-feature-table.js
```diff
@@ -0,0 +1,71 @@
...
+    },
+  },
+  {
+    id: "feature",
+    title: "Feature",
+    display: ({ source }) => {
+      return (
+        <Link href={source.feature}>
+          {source.feature.split("/").slice("-2")}
```
forresttanaka, line 37: Not sure what this does (I don’t _think_ `.slice()` takes a string), but I think it’d be easier to load all the `feature` objects that the phenotypic feature objects refer to in the HumanDonor `getServerSideProps` function, then pass those feature objects into the table `meta` property so you can find them here and display their `term_id` here.

forresttanaka (commented):
Overall looks really good! I’ll just use the Comment option here since this is a draft PR.

components/common-data-items.js
```diff
@@ -24,16 +24,22 @@ import { formatDate } from "../lib/dates";
...
+          <DataItemLabel>Taxa</DataItemLabel>
+          <DataItemValue>{donor.taxa}</DataItemValue>
+        </>
+      )}
+      {donor.ethnicities?.length > 0 && (
+        <>
+          <DataItemLabel>Ethnicities</DataItemLabel>
+          <DataItemValue>{donor.ethnicities.join(", ")}</DataItemValue>
         </>
       )}
```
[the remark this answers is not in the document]
dougli1sqrd: is this what the `children` are for in `<DonorDataItems>`?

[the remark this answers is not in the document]
dougli1sqrd: ah actually, yeah I think I need it there because ethnicities is supposed to come after taxa

pages/human-donors/[uuid].js
```diff
@@ -91,6 +94,14 @@ export default function HumanDonor({
...
         )}
         <Attribution award={award} lab={lab} />
+        {donor.collections?.length > 0 && (
+          <DataPanel>
+            <DataArea>
+              <DataItemLabel>Collections</DataItemLabel>
+              <DataItemValue>{donor.collections.join(", ")}</DataItemValue>
+            </DataArea>
+          </DataPanel>
+        )}
```
Comment (forresttanaka) [inline pages/human-donors/[uuid].js:104]: You should add `collections` to the `<Attribution>` component instead of below it so that you don’t have the horizontal line between them. At some point we’ll probably need to have the `<Attribution>` component take children so that different pages can add their own parts to the Attribution panel, but we don’t yet know what kinds of things will be in this panel, so we can’t do that yet. _Many_ objects have a `collections` property, so it makes sense to make that a part of `<Attribution>` just like award and lab already are.

pages/human-donors/[uuid].js
```diff
@@ -125,6 +138,27 @@ export async function getServerSideProps({ params, req }) {
...
+    const phenotypeTerms = phenotypeTermsList.reduce((obj, el) => {
+      obj[el["@id"]] = el;
+      return obj;
+    }, {});
+
+    const materializedPhenotypicFeatures = phenotypicFeatures.map((feature) => {
+      const term = phenotypeTerms[feature.feature];
+      feature.feature = term;
+      return feature;
+    });
```
Comment (forresttanaka) [inline pages/human-donors/[uuid].js:160]: I’m allergic to `.map()` loops with side effects. I think, after retrieving all the `phenotypic_features` objects, and all their `feature` objects (and I would dedupe the `feature` objects list before doing the `getMultipleFeatures()` request in case many `phenotypic_feature` objects refer to the same `PhenotypeTerm` object — no idea if that’ll be the case, but just in case). Then just send both these arrays to props without tying them together in any way.

Then the `HumanDonor` component takes both arrays. Pass both arrays to `<PhenotypicFeatureTable>`, and inside that pass the `features` array in the `meta` property to `<SortableGrid>`. Then in the `phenotypicFeaturesColumns[x].display` for the `feature` column, get the meta in the second parameter to that function (see the `<SortableGrid>` documentation Markdown file) and then that `display(source, meta)` function searches the `features` property for the item being displayed. It’ll be OK to do a search of the `features` array for each row — it’s highly unlikely to be a performance bottleneck.

forresttanaka (changes requested):
Also, you can get rid everything related to `health_status_history`. That property doesn’t exist anymore, so you can get rid of that section on the HumanDonor page as well as the `healthStatusHistoryColumns` global.

pages/human-donors/[uuid].js
```diff
@@ -125,6 +100,17 @@ export async function getServerSideProps({ params, req }) {
           filterErrors: true,
         })
       : [];
+    const phenotypicFeatures = donor.phenotypic_features
+      ? await request.getMultipleObjects(donor.phenotypic_features, null, {
+          filterErrors: true,
+        })
+      : [];
+    const phenotypeTermsList = phenotypicFeatures
```
Comment (forresttanaka) [inline pages/human-donors/[uuid].js:108]: `phenotypicFeatures` always evaluates truthy because it gets assigned to the empty array if it couldn’t get anything from the server or didn’t exist in the first place. You can test `phenotypicFeatures.length > 0` here.

forresttanaka (changes requested):
I think this is getting pretty close!

components/phenotypic-feature-table.js
```diff
@@ -0,0 +1,84 @@
...
+notes	"Phenotypic feature of body weight"
+status	"released"
+feature	"/phenotype-terms/NCIT_C92648/"
+quantity	58
+quantity_units	"kilogram"
+schema_version	"1"
+observation_date	"2022-11-15"
+creation_timestamp	"2023-03-06T23:31:20.467925+00:00"
+@id: /phenotypic-features/ae1b4a0b-78e6-af0a-8e6d-c0c9b45905fa/
+*/
```
forresttanaka, line 21: What’s this comment for?

components/search/list-renderer/human-donor.js
```diff
@@ -13,10 +13,21 @@ import {
...
 export default function HumanDonor({ item: humanDonor, accessoryData }) {
   const ethnicities =
-    humanDonor.ethnicity?.length > 0 ? humanDonor.ethnicity.join(", ") : "";
+    humanDonor.ethnicities?.length > 0 ? humanDonor.ethnicities.join(", ") : "";
   const sex = humanDonor.sex || "";
   const title = [ethnicities, sex].filter(Boolean);
   const lab = accessoryData?.[humanDonor.lab];
+  const collections =
+    humanDonor.collections?.length > 0
+      ? `${humanDonor.collections.join(", ")}`
```
forresttanaka, line 22: Technically, you can just use `? humanDonor.collections.join(", ")`. The string-template syntax really only helps for composing strings from multiple elements, but after the `.join()` you only have one element. It doesn’t make any execution difference, but would have saved a little typing.

components/search/list-renderer/human-donor.js
```diff
@@ -13,10 +13,21 @@ import {
...
+      : "";
+  const phenotypicFeatures = humanDonor.phenotypic_features
+    ?.map((path) => {
+      const feature = accessoryData?.[path];
+      const notes = feature?.notes ? feature?.notes : "Amount";
+      return `${notes} ${feature?.quantity} ${feature?.quantity_units}`;
+    })
+    ?.join(", ");
```
forresttanaka, line 30: If you executed the `.map()` you’re guaranteed to have an array (possibly an empty one), so you don’t really need the question mark before the `.join()`. Makes no execution difference though.

If `feature` doesn’t exist, `notes` could get displayed with two spaces after it. It almost certainly makes no difference, but sometimes it causes future problems if something later gets displayed after this. I think I probably would have used `if (feature)` and then display different things for each case since so many things are conditional on `feature`.

pages/human-donors/[uuid].js
```diff
@@ -107,6 +78,10 @@ HumanDonor.propTypes = {
   lab: PropTypes.object,
   // Parents of this donor
   parents: PropTypes.arrayOf(PropTypes.object).isRequired,
+  // Phenotypic Features of this donor with the feature turm embedded
+  phenotypicFeatures: PropTypes.arrayOf(PropTypes.object),
+  // Phenotype terms associted with the above features
```
Comment (forresttanaka) [inline pages/human-donors/[uuid].js:83]: `associated`

forresttanaka (dismissed):
This looks really good! I have some feedback that you can change or leave as is.

forresttanaka (changes requested):
I get an error in the Chrome console when displaying a human donor with a phenotype table.

Otherwise, everything looks good!

Screenshot 2023-03-21 at 16 15 31

forresttanaka (approved):
🆒

components/phenotypic-feature-table.js
```diff
@@ -0,0 +1,69 @@
...
+import SortableGrid from "./sortable-grid";
+
+const phenotypicFeaturesColumns = [
+  {
+    id: "phenotypic_feature_id",
+    title: "Phenotypic Feature",
+    display: ({ source }) => {
+      return (
+        <Link href={source["@id"]}>
+          {source.notes ? source.notes : source["@id"]}
```
hitz, line 15: this is incorrect it needs to use phenotypic_features.feature.term_name plus the values in phenotypic_features

PR #127, opened 2023-03-14
IGVF-492-reconcile-treatment

mingjiecn:


components/search/list-renderer/treatment.js
```diff
@@ -20,9 +20,20 @@ export default function Treatment({ item: treatment }) {
...
+            {treatment.treatment_term_name} {UC.mdash}{" "}
+            {treatment.treatment_type} {UC.mdash} {treatment.amount}{" "}
+            {treatment.amount_units} {UC.mdash} {treatment.duration}{" "}
+            {treatment.duration_units}
+          </SearchListItemTitle>
+        ) : (
+          <SearchListItemTitle>
+            {treatment.treatment_term_name} {UC.mdash}{" "}
+            {treatment.treatment_type} {UC.mdash} {treatment.amount}{" "}
+            {treatment.amount_units}
```
forresttanaka, line 34: You could make just the duration parts conditional to avoid the duplication. You can use fragments (e.g. `<>{stuff}</>{duration && <> {duration stuff}</>}`) if that makes it easier than using string templates.

forresttanaka (approved):
Optional change to reduce code duplication.

PR #131, opened 2023-03-23
IGVF-456-human-genomic-variant

dougli1sqrd:
Human genomic variant object page and search result start

This PR on igvfd goes along with this: https://github.com/IGVF-DACC/igvfd/pull/276

### DIFF
--- a/components/search/list-renderer/human-genomic-variant.js
+++ b/components/search/list-renderer/human-genomic-variant.js
@@ -9,6 +9,14 @@ import {
 } from "./search-list-item";
 import { PropTypes } from "prop-types";
 
+export function humanGenomicVariantTitle(variant) {
+  const first = variant.chromosome
+    ? variant.chromosome
+    : variant.refseq_id || variant.reference_sequence;
+  const titleSeq = [first, variant.position, variant.ref, variant.alt];
+  return titleSeq.join(":");
+}
+
 export default function HumanGenomicVariant({ item: variant }) {
   return (
     <SearchListItemContent>
@@ -18,7 +26,7 @@ export default function HumanGenomicVariant({ item: variant }) {
           {variant.uuid}
         </SearchListItemUniqueId>
         <SearchListItemTitle>
-          {variant.refseq_id ? variant.refseq_id : variant.reference_sequence}
+          {humanGenomicVariantTitle(variant)}
         </SearchListItemTitle>
         {variant.rsid && (
           <SearchListItemMeta>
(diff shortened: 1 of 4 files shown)


components/search/list-renderer/human-genomic-variant.js
```diff
@@ -0,0 +1,36 @@
...
+export default function HumanGenomicVariant({ item: variant }) {
+  return (
+    <SearchListItemContent>
+      <SearchListItemMain>
+        <SearchListItemUniqueId>
+          <SearchListItemType item={variant} />
+          {variant.uuid}
+        </SearchListItemUniqueId>
+        <SearchListItemTitle>
+          {variant.refseq_id ? variant.refseq_id : variant.reference_sequence}
```
forresttanaka, line 21: The equivalent could be `variant.refseq_id || variant.reference_sequence` — “if the first one is falsy, render the second”).

pages/human-genomic-variants/[uuid].js
```diff
@@ -0,0 +1,91 @@
...
+export async function getServerSideProps({ params, req }) {
+  const request = new FetchRequest({ cookie: req.headers.cookie });
+  const variant = await request.getObject(
+    `/human-genomic-variants/${params.uuid}`
+  );
+  const breadcrumbs = await buildBreadcrumbs(
+    variant,
+    "assembly",
+    req.headers.cookie
+  );
```
Comment (forresttanaka) [inline pages/human-genomic-variants/[uuid].js:81]: Normally we build the breadcrumbs only _after_ determining the response was a success, since we don’t need the breadcrumbs otherwise.

forresttanaka (commented):
OK, it looks good! Just a couple comments.

components/search/list-renderer/human-genomic-variant.js
```diff
@@ -0,0 +1,44 @@
...
+        <SearchListItemUniqueId>
+          <SearchListItemType item={variant} />
+          {variant.uuid}
+        </SearchListItemUniqueId>
+        <SearchListItemTitle>
+          {humanGenomicVariantTitle(variant)}
+        </SearchListItemTitle>
+        {variant.rsid && (
+          <SearchListItemMeta>
+            <div key={variant.rsid}>{variant.rsid}</div>
```
forresttanaka, line 33: You don’t need a key here since `SearchListItemMeta` is a simple wrapper — not a hidden loop. But also I think all you need is `{variant.rsid}` without the `div` to achieve the same thing.

dougli1sqrd: I initially tried it w/out a div, and it seemed like it wanted  it to be a div or html element, but I could be mistaken

forresttanaka (approved):

components/search/list-renderer/human-genomic-variant.js
forresttanaka: Oh! I was totally wrong. I forgot `SearchListItemMeta` _does_ have a hidden loop, so yeah, you need the `div`. You don’t need the key because it only has one item.

PR #134, opened 2023-03-29
IGVF-190-pooled-from

mingjiecn:
…ct page

### DIFF
--- a/components/common-data-items.js
+++ b/components/common-data-items.js
@@ -210,6 +210,8 @@ export function BiosampleDataItems({
   donors = [],
   biosampleTerm = null,
   diseaseTerms,
+  pooledFrom,
+  partOf,
   options = {
     dateObtainedTitle: "Date Obtained",
   },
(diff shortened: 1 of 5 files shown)


components/common-data-items.js
```diff
@@ -264,6 +266,30 @@ export function BiosampleDataItems({
...
+              ))}
+            </SeparatedList>
+          </DataItemValue>
+        </>
+      )}
+      {partOf && (
+        <>
+          <DataItemLabel>Part of Biosample</DataItemLabel>
+          <DataItemValue>
+            <Link href={partOf["@id"]} key={partOf.uuid}>
```
forresttanaka, line 287: You don’t need a key for the link. `<DataItemValue>` is just a simple wrapper. `<SeparatedList>` and `<SearchListItemMeta>` need keys because they have internal loops.

pages/in-vitro-systems/[id].js
```diff
@@ -135,6 +143,12 @@ export async function getServerSideProps({ params, req }) {
           filterErrors: true,
         })
       : [];
+    const pooledFrom = inVitroSystem.pooled_from
```
Comment (forresttanaka) [inline pages/in-vitro-systems/[id].js:146]: Optionally, I would also check the length of `pooled_from` is non-zero so that you don’t request zero objects. Not sure a zero-length `pooled_from` can happen though.

pages/in-vitro-systems/[id].js
```diff
@@ -135,6 +143,12 @@ export async function getServerSideProps({ params, req }) {
...
+      ? await request.getMultipleObjects(inVitroSystem.pooled_from, null, {
+          filterErrors: true,
+        })
+      : [];
+    const partOf = await request.getObject(inVitroSystem.part_of, null);
```
Comment (forresttanaka) [inline pages/in-vitro-systems/[id].js:151]: It doesn’t look like `part_of` is ever required by any biosample-derived type, so this might crash in that case, maybe?

forresttanaka (changes requested):
A couple small changes, but then I think all the `part_of` requests need a check for existence first. For example, the local insert /tissues/IGVFSM0002DDDD/ doesn’t have a `part_of`, as well as several other local inserts.

forresttanaka (approved):
😎

PR #137, opened 2023-04-04
IGVF-533-attribution

mingjiecn:


forresttanaka (changes requested):
This is really cool! Makes those a lot easier going forward. The changes I noted are very small, but since one exposes a weakness in our ESLint configuration, we might as well take care of that here.

forresttanaka (approved):
😎

PR #139, opened 2023-04-06
IGVF-461-TSV-report-download

dougli1sqrd:
Link and icon to download the currently viewed report as a TSV

### DIFF
--- a/components/search-results/download-tsv.js
+++ b/components/search-results/download-tsv.js
@@ -1,6 +1,9 @@
+// node modules
 import { DocumentArrowDownIcon } from "@heroicons/react/24/outline";
-import { API_URL } from "../../lib/constants";
+// components
 import { ButtonLink } from "../form-elements";
+// lib
+import { API_URL } from "../../lib/constants";
 
 export default function DownloadTSV() {
   const currentQuery = window.location.search;
--- a/components/search-results/download-tsv.js
+++ b/components/search-results/download-tsv.js
@@ -3,8 +3,8 @@ import { API_URL } from "../../lib/constants";
 import { ButtonLink } from "../form-elements";
 
 export default function DownloadTSV() {
-  const current = new URL(window.location.href);
-  const link = new URL(`${API_URL}/report.tsv${current.search}`).toString();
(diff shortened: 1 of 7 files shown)


components/search-results/download-tsv.js
```diff
@@ -0,0 +1,13 @@
+import { DocumentArrowDownIcon } from "@heroicons/react/24/outline";
+import { API_URL } from "../../lib/constants";
+import { ButtonLink } from "../form-elements";
+
```
forresttanaka, line 4: At the very least for consistency’s sake, could you break these up into sections like the other files and sort them alphabetically so that once these import lists get very long, it’s easy to find an import without having to do a search? Also describe what the `DownloadTSV` component is for like all the other components.

components/search-results/header.js
```diff
@@ -18,24 +19,30 @@ export default function SearchResultsHeader({
...
+            </>
+          )}
+        </div>
+
         <div className="mb-1 flex items-center gap-1 sm:mb-0">
           <ItemsPerPageSelector />
         </div>
       </div>
-    </div>
+      <div>
```
forresttanaka, line 42: There’s no gap between the pager and the report table below. You can add a `mb-1` to this div to make that gap the same as the gap between the pager and the dropdown above it.

pages/report.js
```diff
@@ -207,6 +207,7 @@ export default function Report({ searchResults }) {
                   onAllColumnsVisibilityChange,
                 }}
               />
+              <div>after search results header</div>
```
forresttanaka, line 210: Need to delete this.

forresttanaka (changes requested):

forresttanaka:
@dougli1sqrd Also, could you rebase this branch? I’ve merged a couple other tickets and this branch is out of date now.

forresttanaka (dismissed):
Yeah, this looks great! If you want to fix the add.cy.js in this branch, that’ll be fine. I hope there _is_ a fix.

forresttanaka (approved):
OK, nice. Hopefully that Cypress change works reliably, but only one way to find out.

PR #140, opened 2023-04-07
IGVF-412 Disable guest logins

forresttanaka:
Before, I ignored server failure logins, so that meant I let the preceding successful auth0 login stay. I now log out of auth0 if the server login fails and redirect to the /auth-error page.

I also reformatted the error pages so that they all look the same. I had put some error messages in a data panel, while others were the default NextJS ones. I’m trying to move to one that’s more like the NextJS ones, but I did customize those with the new `<Error>` component because the NextJS ones don’t degrade on mobile very well. This custom one is inspired by the default NextJS one but is formatted vertically always instead of horizontally so they stay the same on mobile and desktop:

Pasted Graphic 2

I think the default NextJS one might still appear from time to time, but I’m not sure why (or even if that actually happens; I just thought I saw that happen locally once).

### DIFF
--- a/cdk/infrastructure/config.py
+++ b/cdk/infrastructure/config.py
@@ -52,7 +52,6 @@ config: Dict[str, Any] = {
                 'desired_count': 1,
                 'max_capacity': 4,
             },
-            'backend_url': 'https://igvfd-dev.demo.igvf.org',
             'tags': [
                 ('time-to-live-hours', '72'),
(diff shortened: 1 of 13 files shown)


forresttanaka:
The cdk-check-config-on-pr failure happens because I modified the cdk config since we can’t start igvfd demos. I’ll modify it back before merge.

dougli1sqrd (dismissed):

PR #143, opened 2023-04-12
IGVF-500-targeted-sample-term

mingjiecn:


pages/in-vitro-systems/[id].js
```diff
@@ -31,6 +32,7 @@ export default function InVitroSystem({
   source = null,
   treatments,
   pooledFrom,
+  targetedSampleTerm,
```
Comment (forresttanaka) [inline pages/in-vitro-systems/[id].js:35]: Since `targetedSampleTerm` isn’t required, assign a default value here.

pages/in-vitro-systems/[id].js
```diff
@@ -1,5 +1,6 @@
 // node_modules
 import PropTypes from "prop-types";
+import Link from "next/link";
```
Comment (forresttanaka) [inline pages/in-vitro-systems/[id].js:3]: Could you sort the imports within the section (node_modules in this case) alphabetically by module name (so “n” before “p” in this case)? I find myself sorting these in my tickets, so might as well sort these to begin with. It makes a specific import a lot easier to find in a long list if they’re sorted within their sections.

forresttanaka (changes requested):
Looks nice. Just a couple very small changes.

forresttanaka (approved):
🆒

PR #145, opened 2023-04-13
IGVF-490-reconcile-sample

dougli1sqrd:


forresttanaka (changes requested):
Nice! I just have a couple small changes for consistency and readability. I’m not sure what embedding issue Ben is talking about, unless he means IGVF-554, which is more of an inclusion issue than an embedding one.

components/common-data-items.js
```diff
@@ -13,7 +13,7 @@ import AliasList from "./alias-list";
 import { DataItemLabel, DataItemValue } from "./data-area";
 import DbxrefList from "./dbxref-list";
 import SeparatedList from "./separated-list";
-import SourceProp from "./source-prop";
+import ProductInfo from "./product-info";
```
forresttanaka, line 16: Could you move this above SeparatedList to keep these sorted by module name?

pages/in-vitro-systems/[id].js
```diff
@@ -12,6 +12,7 @@ import {
   DataItemValue,
   DataPanel,
 } from "../../components/data-area";
+import BiomarkerTable from "../../components/biomarker-table";
```
Comment (forresttanaka) [inline pages/in-vitro-systems/[id].js:15]: Could you move this above Breadcrumbs to keep them sorted? This also applies to similar changes to other file under pages/.

forresttanaka (changes requested):
I noticed some things I missed the first time. The new changes look good.

forresttanaka (approved):
Good stuff! I didn’t realize so much code used double equals, so I’m glad we have that rule in the ESLint config now.

PR #155, opened 2023-04-25
IGVF-539-human-donor-phenotype

dougli1sqrd:
Just a note this does not address the bad phenotypic feature table for human donor, as the table itself is fixed in https://igvf.atlassian.net/browse/IGVF-445. This PR/ticket is for human donor search meta showing the term name properly.


pages/human-donors/[uuid].js
```diff
@@ -94,11 +94,7 @@ export async function getServerSideProps({ params, req }) {
       : [];
     const phenotypeTermsList =
       phenotypicFeatures.length > 0
-        ? await request.getMultipleObjects([
-            ...new Set(
-              phenotypicFeatures.map((phenotype) => phenotype.feature)
-            ),
-          ])
+        ? phenotypicFeatures.map((phenotype) => phenotype.feature)
```
Comment (forresttanaka) [inline pages/human-donors/[uuid].js:97]: Since this doesn’t do an igvfd request anymore, this code should move up to the `HumanDonor` UI component. I think the `getServerSideProps()` code should just request needed data and not just extract existing data, with few if any exceptions.

forresttanaka (changes requested):
Looks good. Just one part I think should move from the NextJS server to the browser code.

pages/human-donors/[uuid].js
Comment (dougli1sqrd) [reply pages/human-donors/[uuid].js:97]: Technically this will get deleted once the rodent donor changes go in which fix the phenotypic feature table, so I was using this as a quick placeholder. I think for now I'll just leave this as is until the other ticket merges.

Comment (forresttanaka) [reply pages/human-donors/[uuid].js:97]: OK.

Comment (forresttanaka) [reply pages/human-donors/[uuid].js:97]: Should we hold on this branch until the rodeo donor gets merged?

forresttanaka (approved):
👍

PR #157, opened 2023-04-27
IGVF-478 Redo the Help page

forresttanaka:
The change to rodent donor was because of a coverage failure. I suggested this change to the rodent donor ticket or maybe the human donor ticket.


forresttanaka:
Oops, wait, hang on. I need to update the Page Editor’s Guide. I’ll set it back to Code Review Ready when that’s done.

mingjiecn (commented):
Do we need to add the icons for each help category shown in specification PDF file? Other than that it looks good to me

forresttanaka:
> Do we need to add the icons…

Oh yeah, Khine and I discussed that early on and she said it didn’t really matter to her if we had icons there or not. I figure we have to encode some way to specify an icon and then put that icon into the repo, so no icons for now. Maybe once we’re pretty comfortable with the categories we’ve chosen, someone can design some icons for each section, and then we have to figure out how to specify that icon for each category page.

mingjiecn (dismissed):

PR #166, opened 2023-05-09
IGVF-125-graph-svg 

mingjiecn:


pages/profiles/graph.svg.js
```diff
@@ -0,0 +1,29 @@
...
+
+export default function GraphSvg() {
+  return (
+    <>
+      <Breadcrumbs />
+      <PagePreamble />
+      <>
+        <DataAreaTitle>Graph</DataAreaTitle>
+        <DataPanel>
+          <a href="https://api.data.igvf.org/profiles/graph.svg">
```
forresttanaka, line 17: This always goes to the production server to retrieve the svg, but it’d be nice to go to the server attached to whatever UI instance this runs on. You can import `API_URL` from lib/constants.js and compose the URL from that. `API_URL` for local instances holds `http://localhost:3000`.

pages/profiles/graph.svg.js
```diff
@@ -0,0 +1,29 @@
...
+import Breadcrumbs from "../../components/breadcrumbs";
+import PagePreamble from "../../components/page-preamble";
+import { DataPanel, DataAreaTitle } from "../../components/data-area";
+
+export default function GraphSvg() {
+  return (
+    <>
+      <Breadcrumbs />
+      <PagePreamble />
+      <>
```
forresttanaka, line 14: You don’t need the fragments here. Just make `<DataAreaTitle>` and `<DataPanel>` peers of `<Breadcrumbs>` and `<PagePreamble>`.

pages/profiles/graph.svg.js
```diff
@@ -0,0 +1,29 @@
...
+        <DataAreaTitle>Graph</DataAreaTitle>
+        <DataPanel>
+          <a href="https://api.data.igvf.org/profiles/graph.svg">
+            <Image
+              src="https://api.data.igvf.org/profiles/graph.svg"
+              height={5356}
+              width={4703}
+              alt="IGVF graph"
+            />
```
forresttanaka, line 23: I see where you got the width and height, but as they add and modify schemas, the SVG’s inherent height will change and the image might get cut off or squeezed. Sometimes `<Image>` causes more difficulty than it solves for flexible things like this. I’d recommend this instead:
```
const graphUrl = `${API_URL}/profiles/graph.svg`;
…
<a href={graphUrl} className="block">
  <picture className="block">
    <img src={graphUrl} alt="IGVF graph" />
  </picture>
</a>
```
Using `` gets you an ESLint error unless you wrap it in a `<picture>` element.

pages/profiles/graph.svg.js
```diff
@@ -0,0 +1,29 @@
+// node_modules
+import Image from "next/image";
+
```
forresttanaka, line 3: Just to keep this the same as all other files, you don’t need the blank line between sections. Also below, `data-area` sorts before `page-preamble`.

forresttanaka (changes requested):
Yeah, I think that’s a good decision. An alternative would be to load the image in a `getServerSideProps()` as I mentioned in Slack. Not sure if it has any real advantage though. It would render the SVG on the server, for what it’s worth.

forresttanaka (dismissed):
🆒

PR #171, opened 2023-05-15
IGVF-648-authentication-404

forresttanaka:
Error pages (e.g. 404 pages) in NextJS don’t load environment variables for reasons I can’t fathom, and this causes requests to igvfd to fail because requests for session nonsensically go to the igvf-ui instance instead of the igvfd instance. That means we can’t edit nor add objects, nor can we create access keys, until we reload the page.

This branch does three things to get around these problems:
* Adds a new `/data-provider` API endpoint that retrieves the URL the connected igvfd server. With the addition of a new `getObjectByUrl()` method to FetchRequest, we can now get the `session` and `session-properties` objects by using the entire igvfd URL instead of just the path.
* Adds a new `LinkReloadable` wrapper for `<Link>`. If a page that doesn’t load environment variables comes up, it sets a `GlobalContext` property to indicate that the next click should reload the page. `LinkReloadable` uses an `<a>` tag instead of `<Link>` so that the click causes a full page reload, so that NextJS environment variables get loaded.
* Similarly, it adds a `linkReloadable` library function that wraps `router.push` so that the page gets reloaded if the same `GlobalContext` property got set.

That means any links directly clickable from an error page should use `LinkReloadable` instead of `Link`, though you should import it as `Link` — the link-reloadable.js modules uses a default export so you can call the import anything you want, so let’s just make it `Link`. Similarly, any `router.push()` calls directly from an error page should use the `linkReloadable` function. The vast majority of the UI doesn’t have direct access from error pages, so this is mostly not an issue. Navigation is always available, so those need this treatment, as does the logo above navigation.

### DIFF
--- a/pages/_app.js
+++ b/pages/_app.js
@@ -26,6 +26,8 @@ import ViewportOverlay from "../components/viewport-overlay";
 import "../styles/globals.css";
 
 function Site({ Component, pageProps, authentication }) {
+  // Flag to indicate if <Link> components should cause page reload
+  const [isLinkReloadEnabled, setIsLinkReloadEnabled] = useState(false);
   const { isLoading } = useAuth0();
 
   useEffect(() => {
@@ -49,11 +51,16 @@ function Site({ Component, pageProps, authentication }) {
         type: pageProps.pageContext?.type || "",
       },
       breadcrumbs: pageProps.breadcrumbs || [],
+      linkReload: {
+        isEnabled: isLinkReloadEnabled,
+        setIsEnabled: setIsLinkReloadEnabled,
+      },
     };
   }, [
     pageProps.breadcrumbs,
     pageProps.pageContext?.title,
     pageProps.pageContext?.type,
+    isLinkReloadEnabled,
   ]);
 
   return (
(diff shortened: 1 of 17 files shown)


pages/_app.js
```diff
@@ -49,11 +51,16 @@ function Site({ Component, pageProps, authentication }) {
         type: pageProps.pageContext?.type || "",
       },
       breadcrumbs: pageProps.breadcrumbs || [],
+      linkReload: {
+        isEnabled: isLinkReloadEnabled,
+        setIsEnabled: setIsLinkReloadEnabled,
```
dougli1sqrd, line 56: This is a small thing, so I don't think should make or break the PR, but I wonder if we can make a name that's a little more clear what it's doing. Maybe something like `setReloadable`, or could we have a `disableReload()` and `enableReload()`.

dougli1sqrd (approved):
Great! I just had the one comment, but I'd say take it or leave for the sake of expediency.

pages/_app.js
forresttanaka: I think I’ll keep this as is. To me it’s important that _links_ cause a reload, where `setReloadable` seems _less_ specific about what this relates to. This way, you always see something like `linkReload.isEnabled` which reads just about like the English for what this means.

PR #174, opened 2023-05-17
IGVF-340 Eliminate mapprofile requests

forresttanaka:
The main point of this is to make an alternative to the cascade of /mapprofile requests on every page reload. I removed the mapprofile API code and now use the same method encoded uses.

You can tell whether the user can edit or add an object by checking whether it has an `identifyingProperties` property. From this, you can tell that users can’t edit the `AccessKey` type, nor any of the test schemas, so this is a good test copied from encoded. I also removed the `ProfileMap` component.

This caused a wholesale change to the `canEdit()` usage which now takes a schema instead of an item. I added new functions to determine the schema name from an object and from a collection.

While I was at it, I rewrote the schema index.js display to be more easily styled, as the `indentation` property didn’t work, and it took a different code design to make that unneeded.


components/edit-func.js
```diff
@@ -2,16 +2,21 @@
...
-    return item.actions.find((act) => actions.includes(act.name)) !== undefined;
-  }
-  return false;
+/**
+ * Determines whether the given schema represents an object type that people can edit.
+ * @param {object} schema For type being tested for editability
+ * @returns {boolean} True if the schema represents an object type people can edit
+ */
+export function canEdit(schema) {
+  return Boolean(schema.identifyingProperties?.length > 0);
```
dougli1sqrd, line 19: Why does `identifyingProperties` determine if an object with some schema can be editable?

Also, I think with this the Add and Edit buttons now appear no matter if a user is logged in and has Adding/Editing permissions on an object.

If I'm not logged in and I go to the schema list, I see the buttons to add. But clicking them just brings me to an empty JSON editor. Similarly, if I'm not logged in and I got to an object page and edit it, the Save buttun is always greyed out (rightfully). But we probably shouldn't even show the buttons for people who don't have Add/Edit permissions or who aren't logged in.

dougli1sqrd (changes requested):

components/edit-func.js
forresttanaka: > Why does identifyingProperties determine if an object with some schema can be editable?

I don’t know. That was established even before I started in ENCODE.

> If I'm not logged in and I go to the schema list, I see the buttons to add

Dang! That was an obvious one I completely failed to think about. Thanks!

forresttanaka:
@dougli1sqrd I have rebased on dev with the JSON button ticket merged in and fixed a couple conflicts there, and added a couple Jest tests for new functions introduced with the JSON button ticket.

dougli1sqrd (dismissed):

PR #175, opened 2023-05-17
IGVF-549-json

mingjiecn:


components/json-button.js
```diff
@@ -0,0 +1,61 @@
+/**
+ * Button JsonViewLink and ObjectViewLink are used for swicthing views between object view and json format view.
```
forresttanaka, line 2: sp. “switching”

forresttanaka (changes requested):
This is really cool — much nicer than what we have in encode! I was surprised you put `ObjectPageHeader` into `JsonDisplay`, but I also can’t think of a scenario that would cause problems — I was just surprised.

I just have small changes, mostly in documentation.

forresttanaka:
@mingjiecn Also, in case you didn’t see, you need to rebase and fix the conflicts.

forresttanaka (approved):
Cool! The pattern I’ve been using is to use JSDoc comments for regular functions, and just a comment with no `@param` nor `@returns` for React components, because it’s easier to describe React component properties in PropTypes, and all React components return HTML without exception anyway. Just so you know what pattern I’ve been using. I have some exceptions of course, like the default export for a page seems obvious, as does `getServerSideProps()`, neither of which I ever comment.

PR #176, opened 2023-05-18
IGVF-362 Add user impersonation

forresttanaka:
This implements a new page at `/impersonate-user`, which can also take a `filter=searchterm` query string. Users get loaded in `getServerSideProps()` as a regular `/search?type=User` with `filter=` mapped to `searchTerm=`.

I added the `@tailwindcss/container-queries` plugin so we can use the container queries I presented at a dev meeting a month or two ago. We can now use container queries anywhere.

Impersonation is only available to admins, so this code now checks the session-properties object for its `admin` property, and only shows the `/impersonate-user` page and the menu item for it if you’re an admin. `getServerSideProps()` now detects if you’re not an admin and delivers a 404 if you go directly to the `/impersonate-user` page.

### DIFF
--- a/pages/impersonate-user.js
+++ b/pages/impersonate-user.js
@@ -1,0 +1,168 @@
+// node_modules
+import { useRouter } from "next/router";
+import PropTypes from "prop-types";
+import { useContext, useState } from "react";
+// components
+import useDebounceTimer from "../components/debounce-timer";
+import { Button } from "../components/form-elements";
+import NoCollectionData from "../components/no-collection-data";
+import PagePreamble from "../components/page-preamble";
+import SessionContext from "../components/session-context";
+import { TextField } from "../components/form-elements";
+// lib
+import errorObjectToProps from "../lib/errors";
+import FetchRequest, { HTTP_STATUS_CODE } from "../lib/fetch-request";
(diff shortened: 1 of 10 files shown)


pages/impersonate-user.js
```diff
@@ -0,0 +1,168 @@
...
+
+/**
+ * Send a request to the data provider to impersonate the user with the given @id.
+ * @param {string} userId @id of the user to impersonate
+ * @param {object} session Session object from SessionContext
+ * @returns
+ */
+async function requestImpersonateUser(userId, session) {
+  const request = new FetchRequest({ session });
+  return request.postObject("/impersonate-user", { user: userId });
```
mingjiecn, line 25: Why do we need to post here: /impersonate-user? Can we post directly at home page?

forresttanaka: That’s the igvfd endpoint for triggering impersonation. You post an object with the user’s `@id` to /impersonate-user. Posting anything to "/" wouldn’t do anything.

mingjiecn: Oh, got it. Thanks!

mingjiecn (dismissed):

PR #180, opened 2023-05-25
IGVF-374 Search output page

forresttanaka:
* The two main new components are `<SiteSearchTrigger>` and `<SiteSearch>`. The former is the button that looks like an input field that triggers the search input to appear overlaid on whatever page the user views. The latter is the page that shows site-wide search results.
* Some list view renderers assumed accessory data is always available — an assumption that can cause crashes — which happened as I tested this branch. I fixed these, and explicitly set a null default value on all `accessoryData` properties as they’re not required properties.
* SessionContext now caches the /collection-titles/ mapping, useful for mapping collection names and `@type` to a human-readable name.

### DIFF
--- a/components/site-search-trigger.js
+++ b/components/site-search-trigger.js
@@ -1,0 +1,250 @@
+// node_modules
+import { MagnifyingGlassIcon, XCircleIcon } from "@heroicons/react/20/solid";
+import { Dialog } from "@headlessui/react";
+import { useRouter } from "next/router";
+import PropTypes from "prop-types";
+import { useEffect, useRef, useState } from "react";
+// components
+import { useSessionStorage } from "./browser-storage";
+// lib
+import { KC, UC } from "../lib/constants";
+import { encodeUriElement } from "../lib/query-encoding";
+
+// Maximum number of recent terms to keep track of.
+const MAX_RECENT_TERMS = 5;
+
+/**
+ * Hook to process the array of recently searched terms stored in sessionStorage. The most recent
+ * term gets stored first in the array.
+ */
+function useRecentSearches() {
+  const [recentTermsJson, setRecentTermsJson] = useSessionStorage(
+    "recent-site-search",
+    JSON.stringify([])
+  );
+  const recentTerms = JSON.parse(recentTermsJson);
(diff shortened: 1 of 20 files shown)


components/site-search-trigger.js
```diff
@@ -0,0 +1,250 @@
...
+    const updatedRecentTerms = recentTerms.includes(recentTerm)
+      ? recentTerms.filter((term) => term !== recentTerm)
+      : recentTerms;
+
+    // Add the new term to the front of the list, and limit the length of the list -- older items
+    // fall off the bottom of the list.
+    updatedRecentTerms.unshift(recentTerm);
+    const limitedRecentTerms = JSON.stringify(
+      updatedRecentTerms.slice(0, MAX_RECENT_TERMS)
+    );
```
dougli1sqrd, line 39: Should there be a way to clear this list manually? Maybe not worth worrying about for this particular PR, but was something I thought of

components/site-search-trigger.js
```diff
@@ -0,0 +1,250 @@
...
+    const trimmedTerm = searchTerm.trim();
+    if (trimmedTerm) {
+      saveRecentTerm(trimmedTerm);
+      router.push(`/site-search?term=${encodeUriElement(trimmedTerm)}`);
+    }
+  }
+
+  // Called when the user types a key while the search input is open, allowing ESC to close the
+  // modal, or RETURN to close the modal and trigger the search.
+  function onKeyDown(event) {
```
dougli1sqrd, line 134: I'm curious if we also want a button or something in the modal that performs the search also, like pressing Enter does. But I think that's kinda old-school maybe? So maybe we don't care about that.

dougli1sqrd: ohhh wait nevermind, the magnifying glass is the search button.

dougli1sqrd (approved):

components/site-search-trigger.js
forresttanaka: I don’t know actually. It crossed my mind to put an X next to each to clear individual items, but I couldn’t explain to myself why that would be useful.

PR #184, opened 2023-05-30
IGVF-691 Fix anchor tag documentation

forresttanaka:
We won’t need a demo for this. Just need to make sure I updated not only the descriptions of anchor tags correctly, but also my use of them in other Markdown documents.


khzlin (approved):
this looks good to me

PR #187, opened 2023-05-31
IGVF-683 file table

forresttanaka:
This is the same code Jennifer QAed, so if you pass this, I’ll just merge then.


mingjiecn (approved):

forresttanaka:
@mingjiecn Thank you! I see now how I got confused so I’ll try not to let that happen again.

PR #188, opened 2023-05-31
IGVF-625 Upgrade igvf-ui container to node 18

forresttanaka:


cdk/README.md
```diff
@@ -1,5 +1,5 @@
 ## Infrastructure
-Install Node.js 16 using `nvm` (Node Version Manager):
+Install Node.js 18 using `nvm` (Node Version Manager):
```
keenangraham, line 2: Are we changing the local Node version the CDK uses to deploy infrastructure too, not just the application Node version? This is probably fine but should do the same thing in igvfd so the versions are consistent.

package-lock.json
```diff
@@ -10533,7666 +10549,5 @@
         "url": "https://github.com/sponsors/colinhacks"
       }
     }
-  },
-  "dependencies": {
```
keenangraham, line 10537: Is it expected that a huge chunk of this file isn't needed anymore?

cdk/README.md
forresttanaka: Node 16 becomes unsupported on [September 11](https://endoflife.date/nodejs), so I think it makes sense to switch everything over now.

PR #190, opened 2023-05-31
IGVF-707 Fix display of properties with value of zero

forresttanaka:
We’ve been using `{numericProperty && <Display>}` for several things, but this breaks if the numeric property has the value 0. The result of this expression is the value “0” instead of nothing at all. So I made a new very minimal function (`truthyOrZero()`) we can use to check a numeric value in which 0 is a valid, displayable value.

I made this a function so we don’t have to repeat the same expression twice wherever we need it. An alternative is comparing the typeof the property with the string “untitled” but that has more complicated syntax.

### DIFF
--- a/components/common-data-items.js
+++ b/components/common-data-items.js
@@ -16,6 +16,7 @@ import ProductInfo from "./product-info";
 import SeparatedList from "./separated-list";
 // lib
 import { formatDate } from "../lib/dates";
+import { truthyOrZero } from "../lib/general";
 
 /**
  * Display the data items common to all donor-derived objects.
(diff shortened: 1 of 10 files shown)


dougli1sqrd (changes requested):
This is great! I'm looking through the schemas and cross-referencing in the UI, and I see that in the tissues/ page, the `pmi` field is an optional numeric property:
```
{tissue.pmi && (
                  <>
```
etc, line 61.

I think that's the only one, it's the only one I could find anyway.

forresttanaka:
> I see that in the tissues/ page, the pmi field is an optional numeric property

I saw that, but it has a minimum value of 1, and you get an error if you try to post a 0 value to it. So I figured this one was fine to leave as is.

PR #193, opened 2023-06-06
IGVF-334-json-editor-darkmode

dougli1sqrd:
wiring dark mode status through the global context and making the json editor respond to changes in dark mode.

This code does successfully update the json editor to a dark version of the theme if Mac system dark mode is selected, however the update occurs only after the page is refreshed. So it seems like we're close to what we're looking for. I'd love some guidance on getting this to automatically update the component though.


forresttanaka (approved):
I think that’s a good start. It makes me think we might need to have `DarkModeManager`’s constructor take a callback that gets called whenever the system changes dark mode. Then your callback in `App` sets the `isDarkMode` state.

Can we get rid of the `misspelled option` console warnings? Do we need those options if ace doesn’t even like them?

dougli1sqrd:
I'll work on coveralls

forresttanaka (dismissed):
This is super cool! A surprising number of wranglers seem to use dark mode all the time, so I’m sure they’ll appreciate this. Once you get coveralls back, just set it to QA Ready; I don’t need to review that part.

PR #200, opened 2023-06-13
IGVF-738-sample-source

dougli1sqrd:
source object is now included directly in the TechnicalSample props and loaded on the server


forresttanaka (approved):
I just pushed a change to our Cypress/CircleCI config that so far has two passes in two attempts, so with tremendous luck, we’ll have better luck with these.

PR #208, opened 2023-06-26
IGVF-559-perf-testing

dougli1sqrd:
This adds two config options in next.config.js that sets if the site should log to console output the time each fetch request takes and how long `getServerSideProps` takes for each page.

The utility function `logTime` takes a label and a function which will log how long the function takes using the label in the logging. This is used in getServerSideProps.

Then fetch-request can log how long it takes. So when we load an object page we can see both the getServerSideProps total length as well as all the fetch timings.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -3,6 +3,10 @@
  */
 const UI_VERSION = "4.2.0";
 
+const LOG_FETCH_REQUEST_TIME = true;
+
+const LOG_GET_SERVER_SIDE_REQUEST_TIME = true;
+
 module.exports = {
   reactStrictMode: false,
   images: {
@@ -15,6 +19,8 @@ module.exports = {
   },
   serverRuntimeConfig: {
     BACKEND_URL: process.env.BACKEND_URL || "",
+    LOG_FETCH_REQUEST_TIME,
+    LOG_GET_SERVER_SIDE_REQUEST_TIME,
   },
   publicRuntimeConfig: {
     SERVER_URL: process.env.SERVER_URL || "",
--- a/pages/alignment-files/[id].js
+++ b/pages/alignment-files/[id].js
@@ -27,7 +27,6 @@ import FetchRequest from "../../lib/fetch-request";
 import { isJsonFormat } from "../../lib/query-utils";
 import { logTime } from "../../lib/general";
 
-
 export default function AlignmentFile({
   attribution,
   alignmentFile,
--- a/pages/alignment-files/[id].js
+++ b/pages/alignment-files/[id].js
@@ -25,6 +25,8 @@ import buildBreadcrumbs from "../../lib/breadcrumbs";
 import errorObjectToProps from "../../lib/errors";
 import FetchRequest from "../../lib/fetch-request";
 import { isJsonFormat } from "../../lib/query-utils";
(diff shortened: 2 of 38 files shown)


forresttanaka:
@dougli1sqrd Could you resolve the conflicts and get the tests to pass?

dougli1sqrd:
Yeah, tests were a stub for the moment, wanted to see how you'd like the rest of it in the meantime

next.config.js
```diff
@@ -3,6 +3,10 @@
  */
 const UI_VERSION = "4.2.0";
 
+const LOG_FETCH_REQUEST_TIME = true;
+
+const LOG_GET_SERVER_SIDE_REQUEST_TIME = true;
```
forresttanaka, line 8: So these are just temporary instead of using environment variables, I assume. I see the timings when I try it.

forresttanaka (changes requested):
Overall thoughts, I think this isn’t a bad way to do this. I do wish we didn’t have to wrap every `getServerSideProps` in an IIFE. Most alternatives I can think of are no better, but I do wonder if we could just have an opening statement at the top of the function and then a closing statement before the return for timing. node supports [async localstorage](https://nodejs.org/api/async_context.html) which I’ve never used, but could possibly make that possible.

next.config.js
dougli1sqrd: oh I'm not sure I understand. I hadn't intended originally to use env variables to control this, but we could I guess.

Temporary in what way? Like we would remove these at some point in the future?

pages/alignment-files/[id].js
```diff
@@ -111,46 +113,56 @@ AlignmentFile.propTypes = {
...
-        breadcrumbs,
-        attribution,
-        referenceFiles,
-        isJson,
-      },
-    };
-  }
-  return errorObjectToProps(alignmentFile);
+  return logTime(
+    `/signal-files/${params.id}/`,
```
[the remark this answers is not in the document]
Comment (dougli1sqrd) [reply pages/alignment-files/[id].js:117]: oh I'll use that instead then

next.config.js
forresttanaka: I think environment variables would be good, if they can be set up in a demo. If this goes into dev, I just didn’t want these enabled by default, and have every demo and local instance logging times.

dougli1sqrd:
As compared to the async local storage, just manually doing at the top `const start = performance.now()` and at the end logging `performance.now() - start` for each function seems to be about the same amount of work?

forresttanaka:
> As compared to the async local storage:

Yeah, though I thought it might be nice to hide the console logs and things. It wouldn’t hide much, but it might look a little nicer.

next.config.js
dougli1sqrd: Is it possible to make these false by default and then enable them selectively for a demo or whatever? Also I can just make these env vars. I'm not sure how I'd enable them on a demo though.

forresttanaka: I’m not sure how much value my suggestion has, now. I don’t _think_ you can just deploy a demo with different environment variables, though I could be wrong. So if you have to change the branch to change environment variables (like in a docker file), then that’s no better than just setting these variables in Javascript in a branch.

PR #209, opened 2023-06-26
IGVF-715-use-query

dougli1sqrd:
updating site-search page to use the backend `query` parameter instead of `searchTerm`

### DIFF
--- a/components/docs/search-list-item-renderers.md
+++ b/components/docs/search-list-item-renderers.md
@@ -9,7 +9,7 @@ These components render each of the items that appear when you go to the `/searc
 The results of this search only contain Measurement set objects. However:
 
 ```
-/search?searchTerm=cherry
+/search?query=cherry
 ```
 
 The results can contain a mix of types, so consider each search-result item as standing independently from the others.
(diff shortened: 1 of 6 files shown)


keenangraham:
@forresttanaka would know better but I think there are a couple other places where `searchTerm` is being used, including tests:

https://github.com/IGVF-DACC/igvf-ui/blob/db15e05bb6a965c5ca01e6909cc3b17484f37be2/pages/impersonate-user.js#L53
https://github.com/IGVF-DACC/igvf-ui/blob/db15e05bb6a965c5ca01e6909cc3b17484f37be2/pages/impersonate-user.js#L155
https://github.com/IGVF-DACC/igvf-ui/blob/db15e05bb6a965c5ca01e6909cc3b17484f37be2/cypress/e2e/site-id-search.cy.js#L35-L43
https://github.com/IGVF-DACC/igvf-ui/blob/db15e05bb6a965c5ca01e6909cc3b17484f37be2/components/docs/search-list-item-renderers.md?plain=1#L12

etc.

We should replace any mention of backend `searchTerm` here since it will be removed, and maybe switch over frontend `term` to `query` to be consistent (?) but that could be another ticket.

forresttanaka (changes requested):
Keenan has it right. It’d be nice to change the comments and tests to make future people less confused. Also the impersonation feature uses searchTerm to filter users. The list and report buttons on the site-search page also use searchTerm. In my quick view earlier, I missed all those.

forresttanaka (approved):
I think that’s everything. There’s one comment in impersonate-user, but change it if you think it’s worth it, or just let it go.

PR #220, opened 2023-07-05
IGVF-657 Introduce custom facet mechanism

forresttanaka:
I implemented custom TermLabel and Title components in this ticket. I couldn’t find an existing facet that I could see that could work with a custom Terms facet.

Going forward, we need to track these things:

1. Does the custom Terms component have enough flexibility for all kinds of facets?
2. Do all the custom components get enough data to generate all we need them to do?
3. Do we need a mechanism for custom facet tags?

I can easily see #3 becoming a requirement in the future, but I think we need to implement a custom facet Term component and see how the existing facet tags don’t meet those needs. Then we can come up with a mechanism for custom facet tags.


components/facets/custom-facets/audit-title.js
```diff
@@ -0,0 +1,34 @@
...
+// components
+import { auditMap } from "../../audit";
+
+/**
+ * Displays the title for audit facets. These show a shortened title and an icon matching the icons
+ * in the audit panels. This component gets used for the titles of all audit categories.
+ */
+export default function AuditTitle({ facet }) {
+  // Extract the audit type from the facet field name, and use it to get the color and title.
+  const auditType = facet.field.split(".")[1];
```
pedrohr, line 12: will audit names always have a `.`?

forresttanaka: So far. It was in all of encoded time. I could see the argument for not trusting that though.

pedrohr (dismissed):
just left one question about audit name format, if that's ok, all looks good 💯

components/facets/custom-facets/audit-title.js
pedrohr: if the format is the same then it's all good for now

forresttanaka:
@pedrohr Thanks for the quick review! Yeah, given the encode history, I feel confident about the name format not changing.

PR #221, opened 2023-07-07
IGVF-851 Log requests to NextJS and igvfd servers

forresttanaka:


lib/fetch-request.js
```diff
@@ -124,6 +124,17 @@ const NETWORK_ERROR_RESPONSE = {
 };
 Object.freeze(NETWORK_ERROR_RESPONSE);
 
+/**
+ * Log a request from the NextJS server to igvfd.
+ * @param {string} method FetchRequest method that performs the request
+ * @param {string} path Path or paths to requested resource
+ * @returns {void}
+ */
+function logRequest(method, path) {
```
keenangraham, line 133: Do we want to be able to turn this on/off with a config value? I'm a little worried this will fill logs quickly.

keenangraham (approved):
Looks good. Should determine if we really want all of this in QA.

lib/fetch-request.js
forresttanaka: What config can we change? I think you mean so that we can turn on logging in one demo while all the others don’t. The only config I know how to change is in the repo, which I’d guess is not what you mean.

keenangraham: Maybe in the NextJS config? The simplest thing to do would be to add a constant value in the same file, and switch on it in the `logRequest` fuction (if constant is true console.log, else do nothing). The idea would be to have one value that we can flip to turn on/off all the nonmiddleware logging. In any case maybe not worth worrying about for now.

PR #227, opened 2023-07-18
update version 6.0.0

gabdank:


forresttanaka:
Are we bumping the major version because of the Model pages?

PR #232, opened 2023-07-20
IGVF-809-add-virtual-to-donor-pages

jenjou:


forresttanaka (dismissed):
🆒

components/common-data-items.js
```diff
@@ -47,6 +47,12 @@ export function DonorDataItems({ item, parents = null, children }) {
           <DataItemValue>{item.sex}</DataItemValue>
         </>
       )}
+      {item.virtual && (
+        <>
+          <DataItemLabel>Virtual</DataItemLabel>
+          <DataItemValue>{item.virtual ? "True" : "False"}</DataItemValue>
```
forresttanaka, line 53: Since the only value it can show is “True,” you can just put the string `True` there: `<DataItemValue>True</DataItemValue>`.

forresttanaka (changes requested):

forresttanaka (approved):
👍

PR #233, opened 2023-07-21
IGVF-880-add-nginx-to-frontend

keenangraham:


keenangraham:
@forresttanaka can you review the non-infrastructure, non-nginx changes here?

forresttanaka (approved):
Cool. Why the switch from alpine to bookworm?

keenangraham:
> Cool. Why the switch from alpine to bookworm?

We've never seemed to have issue with it, but there does seem to be decent amount of Alpine Linux hate, where people have had confusing issues with it in production. I also don't think there was a good reason that we used it in first place. This makes it consistent with the distro we use in backend at least: https://github.com/IGVF-DACC/igvfd/blob/ea81c15f7766e6b1dba152e22b7e8bd5a0bc1908/docker/pyramid/Dockerfile#L1

ottojolanki (approved):

PR #235, opened 2023-07-22
IGVF-907-enable-trailing-slashes

keenangraham:


forresttanaka (approved):
Amazing how much that one setting takes care of.

PR #243, opened 2023-08-01
IGVF-856 Improve page-request performance

forresttanaka:
The `getMultipleObjectsBulk()` method of the `FetchRequest` class comprises the main change in this branch. This method replaces `getMultipleObjects()` in most cases for performance. Instead of making individual requests for multiple objects, it instead performs a `/search` request and specifies the `@id`s of each requested object, as well as the fields we want. Because this can result in _very_ long search URLs, these requests can get broken up into multiple parallel requests that fit into the 4094-character URL limit by a generous margin.

This applies mostly to object pages that need an array of objects, like an array of documents or samples. It also applies to search-list pages that might need several objects to render several search-list items that need linked objects. This caused a rewrite to the `getAccessoryDataPaths()` mechanism.

This branch also fixes the case where the user makes a large search or report request by manually editing the `limit=x` query-string parameter. If the user sets the limit to greater than 1000, we redirect to a search page without the `limit=x` parameter. Limits of 1000 or less get retained.

* I’ve moved many bulk requests to their own request functions in the new common-requests.js library file. This lets us have only once place to change the requested fields for particular objects.
* I often replace `uuid` with `@id` as the keys to rendering loops to avoid requesting the uuid in the new bulk requests — not really an issue to request `uuid`, but might as well avoid requesting a field just to use as a key when we can already use the `@id`.
* I removed the `getAndEmbedCollectionObjects()` method from FetchRequest, because we no longer use it.
* I removed the `parent` property of donor objects because they no longer exist in the schemas.
* With `getMultipleObjectsBulk()`, instead of individual objects potentially failing and needing to get filtered out, entire ranges of objects can fail to get retrieved, so keep that in mind.

### DIFF
--- a/lib/fetch-request.js
+++ b/lib/fetch-request.js
@@ -380,7 +380,7 @@ export default class FetchRequest {
    * parallel, it instead requests a `/search`, passing in the `@id` of every requested object, as
    * well as the fields needed for each object.
    * @param {Array<string>} paths Path of each object to request
-   * @param {<any> defaultErrorValue Value to return if the request fails; error object if not given
+   * @param {<any>} defaultErrorValue Value to return if the request fails; error object if not given
    * @param {Array<string>} fields Properties of each object to retrieve
    * @returns {Array<object>} Array of requested objects
    */
@@ -401,7 +401,7 @@ export default class FetchRequest {
       const results = await Promise.all(
         pathGroups.map(async (group) => {
           const pathQuery = group.map((path) => `@id=${path}`).join("&");
-          const query = `${fieldQuery}&${pathQuery}`;
+          const query = `${fieldQuery ? `${fieldQuery}&` : ""}${pathQuery}`;
           const response = await this.getObject(
             `/search/?${query}&limit=${group.length}`,
             null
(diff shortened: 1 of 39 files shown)


lib/fetch-request.js
```diff
@@ -338,44 +376,51 @@ export default class FetchRequest {
...
+            null
+          );
+          if (response?.["@graph"]) {
+            // Add the results to the array of results.
+            return response["@graph"];
+          }
+
+          // The response was an error. Add an empty array to the path groups, so that segment of
+          // results gets skipped.
+          return [];
```
dougli1sqrd, line 416: I wonder if this is what we want to do in case of an error. It looks like we just swallow the error if there is any?

I see a few different ways to deal with errors here:
1. We could collect valid results along with errors. I'll pseudo-code the types like rust, it's how I often think about types haha (Vec is vector, like list or a growable array). So on line 401, `results` would have type: `Vec<Result<Vec<T>, Error>`, and then the consumer would decide how to sort through them. But maybe if we had enough info in the error returned (like what `@id`s was being fetched when some error occurred, we could kinda handle it ourselves, maybe like:
2. `getMultipleObjectsBulk` would have return type: `(Vec<T>, Vec<E>)` collecting all the valid results and all errors into a tuple. Then callers could see what (if any) calls erred and what the error was. But they'd still just have the bulk valid results if the caller wanted to ignore the error.
3. It seems like we don't want this, but I guess we could bail on the entire operation, and have a return type: `Result<Vec<T>, E>` where E is the first error to occur?

dougli1sqrd: In a discussion, we decided we do actually want to do 3, so I'll say "request changes" to reflect that.

dougli1sqrd (changes requested):
We'll change fetch multiple bulk to bail with `[]` at the first error

dougli1sqrd (approved):

PR #249, opened 2023-08-03
IGVF-798-download-button-update

dougli1sqrd:
* Download button is changed to use `DownloadLink` like in some of our other buttons
* Changes `DownloadLink` to allow a disabled state. This is done by just using the same "disabled" css classes as used in a standard button when disabled, however DownloadLink uses a `<div>` to wrap the children components (so it's not a button). But still appears "greyed" out.

### DIFF
--- a/components/form-elements/button.js
+++ b/components/form-elements/button.js
@@ -29,18 +29,36 @@ import React from "react";
 const commonButtonClasses =
   "flex items-center justify-center border font-semibold leading-none";
 
+function enabledButtonTypeClass(bType) {
+  return [
+    `bg-button-${bType}`,
+    `border-button-${bType}`,
+    `text-button-${bType}`,
+    `fill-button-${bType}`,
+  ]; //disabled:bg-button-primary-disabled disabled:border-button-primary-disabled disabled:text-button-primary-disabled disabled:fill-button-primary-disabled`
+}
+
+function disabledButtonTypeClass(bType, pseudoClass = "") {
+  const pc = pseudoClass !== "" ? `${pseudoClass}:` : "";
+  return enabledButtonTypeClass(bType).map((c) => `${pc}${c}-disabled`);
+}
+
 /**
  * Background colors for each of the button types.
  */
 const buttonTypeClasses = {
-  primary:
-    "bg-button-primary border-button-primary text-button-primary fill-button-primary disabled:bg-button-primary-disabled disabled:border-button-primary-disabled disabled:text-button-primary-disabled disabled:fill-button-primary-disabled",
-  secondary:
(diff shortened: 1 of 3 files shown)


forresttanaka:
Reviewing…

components/form-elements/button.js
```diff
@@ -29,18 +29,49 @@ import React from "react";
...
+ * @param {String} pseudoClass If specified, then this will be prepended to each class
+ * in the list e.g. for "disabled": "bg-button-secondary-disabled" becomes
+ * "disabled:bg-button-secondary-disabled"`
+ * @returns List of disabled variant of the button tailwind CSS classes
+ */
+function disabledButtonTypeClass(bType, pseudoClass = "") {
+  const pc = pseudoClass !== "" ? `${pseudoClass}:` : "";
+  return enabledButtonTypeClass(bType).map((c) => `${pc}${c}-disabled`);
+}
+
```
forresttanaka, line 58: I think “disabled” doesn’t look disabled enough. I feel like I can still click on it, though others might disagree. Might want to get Meenakshi’s opinion too: 
image
And in dark mode it looks completely enabled:
image

forresttanaka (changes requested):
You _might_ want to look into whether you can do something like `disabled:bg-button-secondary` instead of `disabled:bg-button-secondary-disabled` through the tailwind.config.js, but:
1. Maybe it’s a pain? But it might make this part much less of a pain.
2. It might not be possible at all.
When I did it this way, I didn’t know Tailwind CSS very well. And I guess I still don’t, but I know enough to know it might be possible. But if you don’t feel like getting into the Tailwind config, I won’t hold it against you.

components/form-elements/button.js
dougli1sqrd: Screenshot 2023-08-02 at 10 50 30 PM

oh interesting because this is what it looks like for me

forresttanaka: The list of [Tailwind CSS core plugins](https://tailwindcss.com/docs/configuration#core-plugins) is really handy because it lets you define custom class names in an easy way. I was hoping it might help for the disabled classes, but it doesn’t look like it does. But that’s what I was talking about last night with the “Tailwind config.”

forresttanaka:
A Jest test fails. Also, you need to rebase on dev because this branch crashes pretty quickly with the current igvfd.

forresttanaka (approved):
😎Code seems simpler this way, too.

forresttanaka:
@dougli1sqrd The change @khzlin requested isn’t difficult. After making that change, just put the ticket directly back to QA Ready. Make sure to update the `FileDownload` comment too.

forresttanaka:
If they introduce a new `upload_status` in igvfd that the UI doesn’t know about, the button will appear enabled. Should it appear disabled instead? Not sure how big a deal that is, so up to you if you want to change it. It’d be a temporary situation.

forresttanaka (approved):
I’ll approve because I think this is fine. But it has an arguable weakness.

PR #251, opened 2023-08-07
IGVF-896-ui-pluralize-3-properties

KAndreeva:
Supported the ui for the following changes from IGVF-895 : 3 properties on many objects will become an array instead of accepting a string: sources, biosample_terms and modifications. Properties are also all being converted from singular to plural naming.


forresttanaka:
Reviewing…

components/product-info.js
```diff
@@ -9,23 +9,23 @@ import { pathToType } from "../lib/general";
 // If no url, then `{source.name} - {productId} - {lotId}`
 // We know if we have a lotId, then we have a productId
 export default function ProductInfo({
-  source = null,
+  sources = null,
```
forresttanaka, line 12: Give this an empty array as a default value (`sources = []`) so that we have no chance of line 17 causing a crash when it tries to use the length of a null property. You could argue to make this a non-optional property, but since conceivably someone else could call `ProductInfo` with a source that might not exist, I think keeping this optional works better.

pages/in-vitro-systems/[id].js
```diff
@@ -37,11 +37,11 @@ import { isJsonFormat } from "../../lib/query-utils";
 
 export default function InVitroSystem({
   inVitroSystem,
-  biosampleTerm = null,
+  sampleTerms = null,
```
Comment (forresttanaka) [inline pages/in-vitro-systems/[id].js:40]: This feedback applies to every file you changed under the "pages" directory. Since `sample_terms` and `sources` are required properties, and embedded, you don’t need to pass the `sampleTerms` and `sources` properties, since they’re already in `inVitroSystem` and the corresponding main properties on the other pages. Just use `inVitroSystem.sample_terms` and `inVitroSystem.sources` directly in this component. You can then also remove both `samplesTerms` and `sources` from PropTypes.

pages/in-vitro-systems/[id].js
```diff
@@ -154,9 +154,8 @@ export async function getServerSideProps({ params, req, query }) {
     `/in-vitro-systems/${params.id}/`
   );
   if (FetchRequest.isResponseSuccess(inVitroSystem)) {
-    const biosampleTerm = inVitroSystem.biosample_term
-      ? await request.getObject(inVitroSystem.biosample_term["@id"], null)
-      : null;
+    const sampleTerms =
+      inVitroSystem.sample_terms?.length > 0 ? inVitroSystem.sample_terms : [];
```
Comment (forresttanaka) [inline pages/in-vitro-systems/[id].js:158]: With the above change, you don’t need to extract `invitroSystem.sample_terms` and `invitroSystem.sources` into their own variables, and you don’t need to return them in `props`.

pages/in-vitro-systems/[id].js
```diff
@@ -204,20 +204,20 @@ export async function getServerSideProps({ params, req, query }) {
...
         partOf,
         targetedSampleTerm,
         biomarkers,
         pageContext: {
-          title: `${biosampleTerm ? `${biosampleTerm.term_name} — ` : ""}${
-            inVitroSystem.accession
-          }`,
+          title: `${
+            sampleTerms.length > 0 ? `${sampleTerms[0].term_name} — ` : ""
+          }${inVitroSystem.accession}`,
```
Comment (forresttanaka) [inline pages/in-vitro-systems/[id].js:220]: Here, you know `sampleTerms` contains something guaranteed, you don’t need to check for a length of zero. Just use `${inVitroSystem.sample_terms[0].term_name} — ${inVitroSystem.accession}`

forresttanaka (changes requested):
🆒 Very nicely done! You took care of a lot of non-obvious changes. I just have changes mostly on the pages that the updated properties from igvfd make possible.

pages/in-vitro-systems/[id].js
Comment (KAndreeva) [reply pages/in-vitro-systems/[id].js:210]: Done

pages/in-vitro-systems/[id].js
Comment (KAndreeva) [reply pages/in-vitro-systems/[id].js]: Done

pages/in-vitro-systems/[id].js
Comment (KAndreeva) [reply pages/in-vitro-systems/[id].js]: Done. Thank you so much for explaining this!

components/product-info.js
KAndreeva: Thank you! I made it an empty array.

pages/tissues/[uuid].js
```diff
@@ -191,15 +181,13 @@ export async function getServerSideProps({ params, req, query }) {
...
-        source,
         treatments,
-        biosampleTerm,
         diseaseTerms,
         pooledFrom,
         partOf,
         biomarkers,
         pageContext: {
-          title: `${tissue.biosample_term.term_name} — ${tissue.accession}`,
+          title: `${tissue.sample_terms[0].term_name} — ${tissue.accession}`,
```
Comment (forresttanaka) [inline pages/tissues/[uuid].js:190]: You can do it this way for the other pages too, like in-vitro system and whole organism. But let’s let that go for now to make sure this makes the release; it doesn’t hurt anything the way it is, but just not necessary.

forresttanaka (dismissed):
😎You can also apply the same change you have in tissues.js to the other pages that use sample_term, but I think we should just leave it to try to get this into the release.

pages/tissues/[uuid].js
Comment (KAndreeva) [reply pages/tissues/[uuid].js:190]: I applied this change to tissues, in vitro system, in whole organism, in technical sample and primary cell.

forresttanaka (approved):

PR #256, opened 2023-08-09
IGVF-791-reconcile-sequence-file

ian-whaling:


forresttanaka:
Starting review…

components/search/list-renderer/file.js
```diff
@@ -26,7 +35,20 @@ export default function File({ item: file }) {
         <SearchListItemMeta>
           <div key="lab">{file.lab.title}</div>
           {summary && <div key="summary">{summary}</div>}
+          {file.dbxrefs && <div key="external resources">{file.dbxrefs}</div>}
```
forresttanaka, line 38: You can comma separate these by using `{file.dbxrefs.join(", ")}`. Right now multiple dbxrefs get squished together.

pages/sequence-files/[id].js
```diff
@@ -71,6 +84,42 @@ export default function SequenceFile({
                     </DataItemValue>
                   </>
                 )}
+                {sequenceFile.flowcell_id && (
+                  <>
+                    <DataItemLabel>Flowcell ID</DataItemLabel>
+                    <DataItemValue>{sequenceFile.flowcell_id}</DataItemValue>
+                  </>
+                )}
+                {truthyOrZero(sequenceFile.lane) && (
```
Comment (forresttanaka) [inline pages/sequence-files/[id].js:93]: This works fine. Looks like `lane` has a minimum value of 1 though, so just using `sequenceFile.lane` without the `truthyOrZero` also works fine. But up to you if you want to change it. I suppose if the schema ever allows a zero value, your method would be more robust, but I doubt we’ll ever have a lane=0?

forresttanaka (changes requested):
😎 Only one very small issue, and one debatable one that’s up to you.

components/search/list-renderer/file.js
ian-whaling: Fixed

pages/sequence-files/[id].js
Comment (ian-whaling) [reply pages/sequence-files/[id].js:93]: Changed, I am also doubtful we’ll ever have a lane=0.

PR #257, opened 2023-08-09
IGVF-921 Convert fetch-request and query library modules to Typescript

forresttanaka:
You’ll also see some changes to the reports modules. That doesn’t relate to Typescript, but I noticed a crash because of an error in a schema. While the schema should fix that, we shouldn’t crash for what’s really a minor schema error.


dougli1sqrd (approved):
I think everything looks good and it's exciting to move towards typescript!

I'm not really sure what to think though about all the optional properties in interfaces like `globals.d.ts`. I'm also trying to figure out idiomatic typescript because it's pretty different than what I'd be used to given other similar type systems.

For example, `Audits` has four optional properties of Array types, and I guess the intention was for this to mirror the json object coming back from the database pretty directly? But we could parse any audit shaped json object into an `Audit` object as having an empty list rather than a missing property. Like `[]` is kind of equivalent to just missing `ERROR` for example.

In my mind we'd define the types as sort of aspirational, and then attempt to parse JSON or objects or whatever into those types. Like to work with Audits we'll need to have extra checks to see if the property exists before we can proceed. It would be easier if we can get to place where we know the layout of Audits (for example) and then operate on that consistent shape.

But maybe the optional properties are just idiomatic, and we have the `?` operator so maybe it's fine? Just saying `X | undefined` feels more explicit to me.

I'm also curious to try out a pattern matching system (https://dev.to/gvergnaud/bringing-pattern-matching-to-typescript-introducing-ts-pattern-v3-0-o1k) that uses destructuring to exhaustively match values to different branches to take. You could have a type `type Foo = X | Y | Z` and then use the match to distinguish a `foo` value among the 3 variants and do something different in each case. It seems more powerful than a switch.

I'm sure we'll iterate on this as we go, so I'll approve this and at least these are my thoughts for now!

PR #258, opened 2023-08-10
IGVF-863-alt-accessions

khzlin:


components/alternate-accessions.js
```diff
@@ -0,0 +1,26 @@
...
+ * // Display the alternate accessions, normally below the header line in objects.
+ */
+export default function AlternateAccession({ alternateAccessions = [] }) {
+  if (alternateAccessions?.length > 0) {
+    return (
+      <h4 className="text-gray-500">
+        {alternateAccessions.length === 1 ? (
+          <span>Alternate Accession: {alternateAccessions[0]}</span>
+        ) : (
+          <span>Alternate Accessions: {alternateAccessions.join(", ")}</span>
```
forresttanaka, line 14: Change both of these from `<span>…</span>` to `<>…</>` to save a bit of HTML. It’ll have the same function but won’t generate the non-necessary spans.

components/alternate-accessions.js
```diff
@@ -0,0 +1,26 @@
+// node_modules
+import PropTypes from "prop-types";
+
+/**
+ * // Display the alternate accessions, normally below the header line in objects.
+ */
+export default function AlternateAccession({ alternateAccessions = [] }) {
```
forresttanaka, line 7: Don’t need the double slash comment; it’s already in a standard comment. Rename this to `AlternateAccessions` since that’s the file name, and it does display potentially more than one.

components/alternate-accessions.js
```diff
@@ -0,0 +1,26 @@
...
+      </h4>
+    );
+  }
+
+  // No alternate accessions to display.
+  return null;
+}
+
+AlternateAccession.propTypes = {
+  alternateAccessions: PropTypes.arrayOf(PropTypes.string), // Array of alternate accession strings
```
forresttanaka, line 25: Move the comment above the property line to be consistent with other components. That gives you more room for a comment too.

components/alternate-accessions.js
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * // Display the alternate accessions, normally below the header line in objects.
+ */
+export default function AlternateAccession({ alternateAccessions = [] }) {
+  if (alternateAccessions?.length > 0) {
+    return (
+      <h4 className="text-gray-500">
+        {alternateAccessions.length === 1 ? (
+          <span>Alternate Accession: {alternateAccessions[0]}</span>
```
forresttanaka, line 12: Should this just say “Alternate: something” instead of “Alternate Accession: something”? I feel like this is really long, and it seems obviously an accession. I know ENCODE uses “Alternate accession.”

components/alternate-accessions.js
```diff
@@ -0,0 +1,26 @@
+// node_modules
+import PropTypes from "prop-types";
+
+/**
+ * // Display the alternate accessions, normally below the header line in objects.
+ */
+export default function AlternateAccession({ alternateAccessions = [] }) {
+  if (alternateAccessions?.length > 0) {
+    return (
+      <h4 className="text-gray-500">
```
forresttanaka, line 10: Should the text be smaller, like on ENCODE? You can add the CSS class `text-sm` or `text-xs` if you think so.

forresttanaka (changes requested):
🆒 A lot of my feedback involves conforming to formatting standards. Plus a couple suggestions if you think they’re a good idea.

components/alternate-accessions.js
khzlin: done

components/alternate-accessions.js
khzlin: i like Alternate Acession better than Alternate.

components/alternate-accessions.js
khzlin: using text-sm

pages/alignment-files/[id].js
```diff
@@ -26,6 +26,7 @@ import { requestDocuments, requestFiles } from "../../lib/common-requests";
 import errorObjectToProps from "../../lib/errors";
 import FetchRequest from "../../lib/fetch-request";
 import { isJsonFormat } from "../../lib/query-utils";
+import AlternateAccession from "../../components/alternate-accessions";
```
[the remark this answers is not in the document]
Comment (khzlin) [reply pages/alignment-files/[id].js:29]: done

components/alternate-accessions.js
```diff
@@ -0,0 +1,26 @@
+// node_modules
+import PropTypes from "prop-types";
+
+/**
+ * // Display the alternate accessions, normally below the header line in objects.
+ */
+export default function AlternateAccession({ alternateAccessions = [] }) {
+  if (alternateAccessions?.length > 0) {
```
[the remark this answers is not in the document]
khzlin: using `.` instead of `?.`

khzlin:
@forresttanaka i think i understand your comment. I changed the file name back to alternate-accessions.js and the component remains AlternateAccessions to match the file property.

forresttanaka:
> @forresttanaka i think i understand your comment. I changed the file name back to alternate-accessions.js and the component remains AlternateAccessions to match the file property.

Yeah. I was just saying both the file name and the component name should both have an “s” at the end, and initially the component name didn’t.

forresttanaka (dismissed):
😎

forresttanaka (approved):
👍

PR #259, opened 2023-08-11
IGVF-812 Construct library object and search pages

forresttanaka:
Most of this PR isn’t novel, but builds on what we already have. But one novel piece involves the collapsable data item values. You get a button that lets the user expand and collapse data areas that might get large. We can use this for existing properties too, but I only used it on ConstructLibrary objects for now, where the ticket specified.

I extracted the MeasurementSet sequencing file table to its own module, so ConstructLibrary can share it.

### DIFF
--- a/pages/construct-libraries/[id].js
+++ b/pages/construct-libraries/[id].js
@@ -185,15 +185,6 @@ function LibraryDetails({ library, clonedGenes, children }) {
       </Reporter>
     );
   }
-  if (specificDetails) {
-    return (
-      <>
-        {specificDetails}
-        <DataItemValue>{library.scope}</DataItemValue>
-        <DataItemLabel>Selection Criteria</DataItemLabel>
-      </>
-    );
-  }
   return null;
 }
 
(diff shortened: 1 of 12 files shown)


pages/construct-libraries/[id].js
```diff
@@ -0,0 +1,450 @@
...
+                <>
+                  <DataItemLabel className="flex items-baseline">
+                    Targeted Loci
+                    <DataItemValueExpandButton
+                      isExpandable={isTargetedLociExpandable}
+                      isExpanded={isTargetedLociExpanded}
+                      onClick={setIsTargetedLociExpanded}
+                    />
+                  </DataItemLabel>
+                  <DataItemValue>
```
Comment (mingjiecn) [inline pages/construct-libraries/[id].js:308]: Do you mean this <DataItemValue> component can expand and collapse? I am not sure how?

Comment (forresttanaka) [reply pages/construct-libraries/[id].js:308]: It appears collapsed by default:
Screenshot 2023-08-11 at 16 19 32

When you click the icon after the label, it expands:
Screenshot 2023-08-11 at 16 20 07

mingjiecn (dismissed):

pages/construct-libraries/[id].js
Comment (mingjiecn) [reply pages/construct-libraries/[id].js:308]: Thank you!

PR #278, opened 2023-08-21
IGVF-1007-configuration-file-pages

ian-whaling:


pages/configuration-files/[...id].js
```diff
@@ -0,0 +1,219 @@
...
+        <JsonDisplay item={configurationFile} isJsonFormat={isJson}>
+          <DataPanel>
+            <DataArea>
+              <FileDataItems
+                item={configurationFile}
+                fileSet={fileSet}
+              ></FileDataItems>
+            </DataArea>
+          </DataPanel>
+          {seqSpecOf?.length > 0 && (
```
Comment (forresttanaka) [inline pages/configuration-files/[...id].js:72]: Not a big thing, but the `seqSpecOf` property is required (might be empty, which satisifies this requirement), so you can just do `seqSpecOf.length` without the question mark.

forresttanaka (dismissed):
Just one tiny feedback that you can fix if you feel like, but I have no objections if you don’t.

forresttanaka (approved):
Nice!

PR #286, opened 2023-08-28
IGVF-274-breadcrumb-values

dougli1sqrd:
change buildBreadCrumbs to accept the breadcrumb title directly rather than property name
update pages to use the correct properties


dougli1sqrd:
Gonna fix tests

PR #287, opened 2023-08-28
IGVF-1040-ui-update-display-file-set-type

KAndreeva:
Updated UI to display file_set_type for model and auxiliary file sets.


forresttanaka (approved):
👍

PR #289, opened 2023-08-30
IGVF-836-ui-add-modification-page-and-search

KAndreeva:
Supported display of cas_species, description and sources (if present), and documents on the modification object page.


forresttanaka (approved):
🆒

PR #291, opened 2023-08-30
IGVF-922-lib-typescript

dougli1sqrd:
* lib/general.ts
* lib/color.ts

to get started

### DIFF
--- a/lib/color.ts
+++ b/lib/color.ts
@@ -1,7 +1,7 @@
 interface Color {
-  r: number,
-  g: number,
-  b: number
+  r: number;
+  g: number;
+  b: number;
 }
 
 /**
--- a/lib/general.ts
+++ b/lib/general.ts
@@ -57,7 +57,7 @@ export function sortedJson(obj: JSON): JSON {
   }
   // We know it's not an array at this point
   if (typeof obj === "object") {
-    const sorted: { [key: string]: JSON } = {} ;
+    const sorted: { [key: string]: JSON } = {};
     const o = obj as { [key: string]: JSON };
     Object.keys(o)
       .sort()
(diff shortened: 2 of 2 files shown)


lib/color.ts
```diff
@@ -7,7 +13,7 @@
  * @return {object.g} Green component as an integer from 0 to 255
  * @return {object.b} Blue component as an integer from 0 to 255
```
forresttanaka, line 14: Change line 11 from `{object}` to `{Color}`.

lib/color.ts
```diff
@@ -23,7 +29,7 @@ export function colorToTriple(color) {
  * @param {string} color CSS hex color to test
  * @return {bool} True if `color` is a "light" color.
```
forresttanaka, line 30: Change to `boolean`. I used to use `bool` because that’s what `PropTypes` uses.

lib/general.ts
```diff
@@ -3,7 +3,7 @@
  * @param {string} path The @id of the object to get the type for.
  * @returns Type of the object this path comes from; or the empty string if no type could be found.
```
forresttanaka, line 4: Hmm. I forgot to have a type here. So `@returns {string}`.

lib/color.ts
```diff
@@ -1,3 +1,9 @@
+interface Color {
```
forresttanaka, line 1: I think to call this function from Typescript, you might need to export this interface. I’m about 60% sure of that. You could try it by converting the color.test.js Jest test to Typescript.

lib/general.ts
```diff
@@ -26,13 +26,23 @@ export function isValidUrl(url) {
...
   return url.split("?")[0];
 }
 
+type JSON =
+  | null
+  | boolean
+  | string
+  | number
+  | Array<JSON>
+  | { [key: string]: JSON };
```
forresttanaka, line 44: I wonder if this should go in globals.d.ts? Not sure if this type is useful outside of this function or not.

forresttanaka (changes requested):
Most of this is about comments. I do have a suggestion about moving types/interfaces around, but I’m really not sure if that’s needed or not. But it’s easy to test by converting the corresponding Jest tests to Typescript.

lib/general.ts
dougli1sqrd: Yeah not sure either. I was going to keep it this way for now, and as we converted more if I needed `JSON` I would move to a better location

lib/color.ts
dougli1sqrd: This looks like it works okay. I'll commit the test as ts as well

forresttanaka:
@dougli1sqrd Lint error in a couple files.

forresttanaka (approved):
😎

PR #296, opened 2023-09-05
IGVF-822 Update MeasurementSet pages

forresttanaka:
I had the mechanism to collapse and expand data-area values before, but because we only used it in one place, this mechanism wasn’t very reusable. This ticket uses that mechanism in more places, so I introduced a custom hook and associated components in data-area.js to display these, as well as the concept of a “collapser” object that tracks this state. I also changed the UI to not have the expand/collapse button right after the label, but instead right after the data — which has its own issues but I think people generally prefer it. The label now includes “more” and “fewer” text as well as a count of hidden items, instead of just an ellipsis icon.

Because we now show an “Associated seqspec files” column on sequencing file tables, and some pages already use this column, I had to request seqspec files on these pages as well.

### DIFF
--- a/pages/configuration-files/[...id].js
+++ b/pages/configuration-files/[...id].js
@@ -36,7 +36,6 @@ export default function ConfigurationFile({
   configurationFile,
   fileSet = null,
   seqspecOf,
-  seqspecFiles,
   sequencingPlatforms,
   documents,
   derivedFrom,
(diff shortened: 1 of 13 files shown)


pages/configuration-files/[...id].js
```diff
@@ -145,9 +149,24 @@ export async function getServerSideProps({ params, req, query, resolvedUrl }) {
...
-      : [];
+
+    // Get any seqspec_of files, as well as their associated seqspec files.
+    let seqspecOf = [];
+    let seqspecFiles = [];
+    if (configurationFile.seqspec_of) {
+      seqspecOf = await requestFiles(configurationFile.seqspec_of, request);
+
+      if (seqspecOf.length > 0) {
+        const seqspecPaths = seqspecOf
```
Comment (khzlin) [inline pages/configuration-files/[...id].js:160]: i'm not sure of seqspecPaths. seqspec_of is reverse linked to file.seqspec so I think seqspec_of.file.seqspec should give the same results? Talked to Forrest to drop Associated Seqspec File from 'seqspec File of' table

dougli1sqrd (approved):

PR #297, opened 2023-09-05
IGVF-1048-linking-on-supplement

ian-whaling:


components/search/list-renderer/analysis-set.js
```diff
@@ -20,7 +21,14 @@ export default function AnalysisSet({ item: analysisSet }) {
...
     analysisSet.input_file_sets?.length > 0
-      ? analysisSet.input_file_sets.map((fileSet) => fileSet.accession)
+      ? analysisSet.input_file_sets
+          .map((fileSet, index) => [
+            <Link key={fileSet["@id"]} href={fileSet["@id"]}>
+              {fileSet.accession}
+            </Link>,
+            index < analysisSet.input_file_sets.length - 1 && ", ",
+          ])
+          .flat()
```
forresttanaka, line 31: You can do this in an easier and more standardized way using the [SeparatedList component](https://github.com/IGVF-DACC/igvf-ui/blob/dev/components/separated-list.js). We have examples all over the code base. You can just collect up all the input file sets and flatten them into one array, then put a map loop within `<SeparatedList>`. Probably easier to do the render part inline with `<SearchListItemSupplementContent>`.

components/search/list-renderer/file.js
```diff
@@ -26,6 +27,17 @@ export default function File({ item: file, accessoryData = null }) {
...
+    file.seqspec_of?.length > 0 && accessoryData
+      ? file.seqspec_of
+          .map((key, index) => [
+            <Link key={key} href={key}>
+              {accessoryData[key].accession}
+            </Link>,
+            index < file.seqspec_of.length - 1 && ", ",
+          ])
+          .flat()
+      : [];
```
forresttanaka, line 40: Same thing as the analysis-set.js file, using `<SeparatedList>`.

components/search/list-renderer/file.js
```diff
@@ -48,16 +60,28 @@ export default function File({ item: file, accessoryData = null }) {
...
+                </SearchListItemSupplementLabel>
+                <SearchListItemSupplementContent>
+                  <Link href={fileSet["@id"]}>{fileSet.summary}</Link>
+                </SearchListItemSupplementContent>
+              </SearchListItemSupplementSection>
+            )}
+            {seqspecOfFileLinks.length > 0 && (
+              <SearchListItemSupplementSection>
+                <SearchListItemSupplementLabel>
+                  Seqspec of
```
forresttanaka, line 78: Technically a capital on the `of`, though I think this is styled as all uppercase anyway.

forresttanaka (changes requested):
Nice! Nothing wrong but I have a suggestion for a tool we already have to make the code easier and more easily recognizable for what it does at a glance.

components/search/list-renderer/analysis-set.js
ian-whaling: Replaced.

components/search/list-renderer/file.js
ian-whaling: Replaced.

components/search/list-renderer/file.js
ian-whaling: Replaced.

PR #300, opened 2023-09-06
IGVF-1013-matrix-file-pages

ian-whaling:


pages/matrix-files/[...id].js
```diff
@@ -0,0 +1,210 @@
...
+          <DataPanel>
+            <DataArea>
+              <>
+                <DataItemLabel>First Dimension</DataItemLabel>
+                <DataItemValue>{matrixFile.dimension1}</DataItemValue>
+              </>
+              <>
+                <DataItemLabel>Second Dimension</DataItemLabel>
+                <DataItemValue>{matrixFile.dimension2}</DataItemValue>
+              </>
```
Comment (forresttanaka) [inline pages/matrix-files/[...id].js:82]: You don’t need the `<>…</>` on these two label/value pairs. React is really confusing on this, but every JSX expression can return a maximum of one component. That one component can wrap many other components. So the return on line 51 needs the `<>…</>` because it otherwise would return both the `<Breadcrumbs>` _and_ `<EditableItem>` components — two is too many. So we wrap both of those in one `<>…</>`. The `<DataArea>` above here on line 74 can be one of many because it’s wrapped by one `<DataPanel>` which itself can be many wrapped by `<JsonDisplay>`, which itself can be one of many wrapped by `<EditableItem>`, which _now_ can be one of many because it’s wrapped by the single `<>…</>`.

The `referenceFiles` expression starts a new JSX expression. You can’t have both `<DataItemLabel>` _and_ `<DataItemValue>` because that’s two. So they get wrapped by one `<>…</>`.

The `<>…</>` you use above don’t help (neither do they hurt, but it’s not standard JSX) because they already have the `<DataArea>` wrapper which can contain as many components as you like.

It’s weird, and the early days of React saw lots of complaints about this, especially since `<>…</>` didn’t exist, so you had to use HTML-generating `<div>` or `<span>` to wrap. But now people are used to it, and `<>…</>` generates no HTML — it just satisfies the syntax. Written Korean has a rule that’s kind of similar, which I found fascinating.

pages/matrix-files/[...id].js
```diff
@@ -0,0 +1,210 @@
...
+            <>
+              <DataAreaTitle>
+                Files {matrixFile.accession} Derives From
+              </DataAreaTitle>
+              <DerivedFromTable
+                derivedFrom={derivedFrom}
+                derivedFromFileSets={derivedFromFileSets}
+              />
+            </>
+          )}
```
Comment (forresttanaka) [inline pages/matrix-files/[...id].js:109]: And to continue to say way too much about this, this is correct — you need this `<>…</>` because this `derivedFrom` loop starts a new JSX expression, so you can’t return both `<DataAreaTitle>` _and_ `<DerivedFromTable>` — two is too many.

forresttanaka (changes requested):
This is great, just some syntax standardization for which I included probably way more discussion than you want to read about, and I didn’t tl;dr it.

PR #301, opened 2023-09-10
IGVF-1018-tooltips

dougli1sqrd:
* tooltip component which will display text content upon being clicked. 
* TooltipButton which is a div styled like a button for use with Tooltip.
* Tooltips were added to the schema page, using the descriptions as the content

~~Code Review note: TooltipButton is maybe not the slickest way to do this, but Popover already renders a button so I thought it would be pretty easy to just make something that renders like a button and use that as the child element to Popover.~~


components/tip.js
```diff
@@ -0,0 +1,36 @@
...
+import { Popover } from "@headlessui/react";
+import { useState } from "react";
+import { usePopper } from "react-popper";
+import PropTypes from "prop-types";
+
+/**
+ * When the children element is clicked a panel comes up on
+ * top of the page with the given text content.
+ */
+export default function Tooltip({ content, children }) {
```
forresttanaka, line 11: Since the default function is called `Tooltip`, could you name the file `tooltip.js`? Makes it easier to predict one from the other when using this.

pages/profiles/index.js
```diff
@@ -91,6 +100,17 @@ function SubTree({ tree, objectType, schemas, collectionTitles = null }) {
...
+              <Button
+                onClick={() => {}}
+                label={`View the summary of the ${title} schema`}
+                type="secondary"
+                size="sm"
+                hasIconCircleOnly
+              >
+                <QuestionMarkCircleIcon />
+              </Button>
+            </Tooltip>
```
forresttanaka, line 113: Seems a little painful to have to use a button that doesn’t do anything everywhere we’ll need a tooltip. Can the `Popover.Button` be styled the way we need it? Or just use `<Button>` inside `<Tooltip>` instead of `<Popover.Button>`?

forresttanaka (changes requested):
The tooltip message is invisible in dark mode. I’d _really_ like not to have the `<Button>` component in every place we need a tooltip. I’d feel really surprised if there was no other way to do this. Just inserting `<Tooltip>` in the place we need it would be optimal. A default icon with the question mark would also be great, and we can nest something else if we wanted something else instead of the question-mark icon.

I think we’ll eventually need to be able to pass a React component as the tooltip content, though not yet necessary. Not hard to do though. If we ever passed in styled text or links, we’d need that.

forresttanaka (approved):
Cool — That’ll be much easier once we use Tooltip in more places. The only other feedback I have is to style the tooltips with smaller text, less line height, and more even margins (and use padding instead of margins). e.g.:

Screenshot 2023-09-10 at 15 54 40

PR #307, opened 2023-09-20
IGVF-1080-typescript

dougli1sqrd:
This converts access-keys, attachment, and attribution to typescript.

In attribution, we have a lot of interactions with `getObject` with `FetchRequest` and need to do some error handling. In JS this is pretty easy, but to get the types to line up well I added an interface `IsError` with just the property `isError: true`. Then to the `ErrorObject` I added this property. This will make it easier to detect if an `ErrorObject` has been returned. Then, because the normal way that `getObject` returns a default value upon error, it actually has 3 possible return types since typescript doesn't know that `getObject` can't return an `ErrorObject` when we pass in a default.

To make this easier, I made a function `nullOnError<T, E extends IsError>(x: T | E | null): T | null` that ensures we return null if `x` is a `IsError`, just like wanted. Since the type system still needs help with the generics, when I call it I manually fill in the types.

Lastly, using something like `IsError` is really only a bandaid fix. I see that there's a `@type` of "Error" on the ErrorObject, and we could use that, but this seemed easier than going through a list. Additionally, at some point if `getObject` fails we should have a proper way to wrap the error value in a Result type that can say that it's a proper error. I think using `IsError` is a little more clear than using the Error `@type` field because `@type` part of the API of the db objects themselves, and it's conceivable that there could be an object that we are legitimately trying to fetch with an Error type, and it wouldnt' be clear if it's a failure of `getObject` or just we legitimately want to fetch that type of object. (Think of In Band vs Out of Band communication)

I could see us wanting to move `IsError` or `nullOnError` though


dougli1sqrd:
looking into the cypress tests

forresttanaka (changes requested):
Mostly I need more documentation, as I don’t know when I’m supposed to use the new functions and types. But also a stray console.log.

forresttanaka (approved):
Nice!

PR #308, opened 2023-09-25
IGVF-378 Add facet to select abstract types

forresttanaka:
The main change was the implementation of a custom facet-terms component — type-terms.js. This is relatively complicated because it has to render a hierarchy, so I have code to map the array of “type=” terms according to the /profiles._hierarchy object.

I also added a new form control for radio buttons that anyone can use for future radio buttons. We might want to make a radio-button group component that tracks the selected button, unneeded for the type facet radio buttons.

I had the concept of a facet key to cause the currently selected facet group to reset. It appears no longer needed, and was resetting the facet group selection when selecting a new type, so I have removed that mechanism.

### DIFF
--- a/components/facets/__tests__/facet-groups.test.js
+++ b/components/facets/__tests__/facet-groups.test.js
@@ -370,31 +370,34 @@ describe("Test the `<FacetGroup>` component", () => {
     expect(facets.length).toBe(2);
 
     // Check that the facets have the predicted order and correct titles.
-    let facetTitle = within(facets[0]).getByRole("heading", { name: /Sex/ });
+    let facetTitle = within(facets[0]).getByRole("heading", { name: /Lab/ });
     expect(facetTitle).toBeInTheDocument();
-    facetTitle = within(facets[1]).getByRole("heading", { name: /Lab/ });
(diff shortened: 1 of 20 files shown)


components/facets/__tests__/facet-groups.test.js
```diff
@@ -370,31 +370,34 @@ describe("Test the `<FacetGroup>` component", () => {
...
+    let facetTermCheckbox = within(facetTerms[1]).getByRole("checkbox");
     expect(facetTermCheckbox).toBeInTheDocument();
     expect(facetTermCheckbox).toHaveAttribute(
       "aria-label",
-      expect.stringMatching(/^female with 6 results$/)
+      expect.stringMatching(/^male with 1 result$/)
     );
     expect(facetTermCheckbox).not.toHaveAttribute("checked");
 
     // Check that the second term has the correct label and is not checked.
```
mingjiecn, line 395: Comment said check the first term, but `facetTerms[1] `seems to be the second term?

forresttanaka: Oops! Correct. It wouldn’t surprise me if you see more, as I kept going back and forth on these.

mingjiecn (dismissed):

PR #312, opened 2023-09-27
IGVF-808-related-donors-table

ian-whaling:


components/related-donors-table.js
```diff
@@ -0,0 +1,41 @@
...
+// components
+import { DataGridContainer } from "./data-grid";
+import SortableGrid from "./sortable-grid";
+
+const relatedDonorsColumns = [
+  {
+    id: "related_donor_id",
+    title: "Donor",
+    display: ({ source }) => {
+      return <Link href={source.donor}>{source.donor.substring(14, 28)}</Link>;
```
forresttanaka, line 13: I wouldn’t extract the accession from the path, especially not by hardcoding character offsets since the slightest (even unlikely) change to the path will break this. Regex would be more maintainable, but still not as good as loading the donor objects and using their `accession` property.

Two ways to do that:
1. Change igvfd to embed at least the `accession` and `@id` properties of `related_donors`. Then this `display` property can just display `source.donor.accession`.
2. Change `RelatedDonorsTable` to not only take the existing `related_donors` property, but also an array of donor objects. Then this `display` property searches this array for the matching `@id` and then displays the matching donor’s `accession` property. To get this, the `HumanDonor` `getServerSideProps` needs to extract an array of `related_donors[x].donor` paths, dedupe them, then call `requestDonors` from the `lib/common-requests` file to retrieve those donors. Pass the resulting array as a new property in the `getServerSideProps` return, then add this as a new property to  the `HumanDonor` page component, then pass that to `RelatedDonorsTable`. We do this on the [In Vitro System](https://github.com/IGVF-DACC/igvf-ui/blob/e03435fc70d32ea624a98791aca33526c3f369dc/pages/in-vitro-systems/%5Bid%5D.js#L210) page, though you have to do a bit more work to extract the donor paths from the `related_donors` array first.

forresttanaka (changes requested):

components/related-donors-table.js
forresttanaka: Personally I feel like 1. is the better option, as it’s easier, and you wouldn’t embed much. The downside of course is you have to open a new igvfd ticket and get that change in first.

components/related-donors-table.js
```diff
@@ -0,0 +1,41 @@
...
+    id: "related_donor_id",
+    title: "Donor",
+    display: ({ source }) => {
+      return <Link href={source.donor["@id"]}>{source.donor.accession}</Link>;
+    },
+  },
+  {
+    id: "relationship_type",
+    title: "Relationship Type",
+    display: ({ source }) => source.relationship_type,
```
forresttanaka, line 19: I doubt you need the `display` property at all. If you don’t have one, it just gets the property matching `id` and displays it, which is all this `display` function does. You need `display` for `related_donor_id` in the first column because the built-in mechanism only displays contents and can’t make a link out of it, but the built-in mechanism can totally handle this one.

forresttanaka (changes requested):
My feedback is truly not a big deal and will result in no change to the rendered page, but I feel like we should not have a function where it’s not needed. At least I _think_ it’s not needed, if I remember how this works well enough.

PR #314, opened 2023-10-02
IGVF-959 Abstract and multiple types for reports

forresttanaka:
* While I changed the page filename from report.js to multireport.js to work with the NextJS router, I decided to keep all the supporting files named “report,” and keep “multireport” limited to anything path related.
* I moved the list and report pager to appear in a new slot between the results count and the results. People were sometimes confused about the space between the other controls and the results before noticing the pager on the right side. I also only show one page on each side of the selected page instead of two on each side, so it works on mobile better. The page numbers also appear narrower, also to work on mobile better.
* I removed the manually entered columns from the column-selection modal. It was an awkward UI.

### DIFF
--- a/components/pager.js
+++ b/components/pager.js
@@ -15,45 +15,45 @@ import { Button } from "./form-elements";
  * component allow the user to move to the previous and next page to the current page. We have four
  * cases to consider based on the total page count:
  *
- * Nine or fewer: Straight sequence of pages with no ellipses ([x] indicates current page):
- * <  1   2   3   4   5  [6]  7   8   9  >
+ * Seven or fewer: Straight sequence of pages with no ellipses ([x] indicates current page):
+ * <  1   2   3   4   5  [6]  7  >
  *
- * More than nine with the current page towards the left end ('.' indicates ellipsis):
- * <  1   2  [3]  4   5   6   7   .   20  >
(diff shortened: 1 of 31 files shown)


components/pager.js
```diff
@@ -15,45 +15,45 @@ import { Button } from "./form-elements";
...
 export default function Pager({ currentPage, totalPages, onClick, className }) {
   // Create the array of 1-based page numbers, with page 0 to represent an ellipsis before the
   // current page and page -1 to represent an ellipsis after the current page. No real difference
   // between these two ellipsis values, but Pager uses distinct values so we don't have duplicate
-  // React keys. No point memoizing `pageNumbers` as it needs recalculating when any prop changes.
+  // React keys.
   let pageNumbers;
-  if (totalPages <= 9) {
-    // A total page count of nine or fewer has no ellipses -- just a straight array of
+  if (totalPages <= 7) {
```
dougli1sqrd, line 45: Maybe overkill, but would pulling this out into a configuration variable be useful? Would be easier than changing all the docs and references to this number in this page if we ended up changing this in the future.

forresttanaka: I would normally want that myself, but as this was a side quest and it would take time to work out all the different parameters (maximums, minimums, margins, etc.), I decided to just modify what we had for now.

mingjiecn (dismissed):

PR #318, opened 2023-10-05
IGVF-1017 Allow group and wildcard facet selections

forresttanaka:
* Update npm packages.
* Prepare for handling buttons in the title.
* Make the wildcard buttons fully functional. Have selections of regular terms remove wildcard selections.
* Update package-lock.json after rebase to dev.
* Add custom facet element for tag labels. Add a standard tag-label as well as one for taxa.
* Implement select-all and select-none buttons.
* Add Clear All button to search results.
* Make new selection controls accessible.
* Full Jest test coverage.
* Update the facet documentation to include the custom facet tag labels.
* Remove a console.log.

### DIFF
--- a/components/long-click.js
+++ b/components/long-click.js
@@ -7,6 +7,11 @@ import { useEffect, useRef } from "react";
 const DEFAULT_DELAY = 250;
 
 /**
+ * Value to use to indicate an unset long-click timer.
+ */
+const TIMER_UNSET = "UNSET";
+
+/**
  * Hook to detect short and long clicks in an element. Pass it the ID of the element to attach the
  * event listeners to. Pass it the regular- and long-click callbacks. It calls the regular-click
  * callback when the user short-clicks the element, and the long-click callback when the user
@@ -33,7 +38,7 @@ export default function useLongClick(
 ) {
   // The "UNSET" timer value indicates that the timer isn't active. That's as opposed to the null
   // value, which indicates that the timer has expired
-  const timer = useRef("UNSET");
+  const timer = useRef(TIMER_UNSET);
 
   /**
    * Called when the user presses the mouse button or touches the touchscreen. Start the long-click
@@ -56,10 +61,10 @@ export default function useLongClick(
       onClick();
     } else {
       // The mouse-up event happened after the click timer expired, so the user long-pressed the
-      // checkbox. Call the onLongChange callback if it exists.
+      // checkbox. Call the onLongClick callback if it exists.
       onLongClick();
     }
-    timer.current = "UNSET";
+    timer.current = TIMER_UNSET;
   }
 
   /**
--- a/components/long-click.js
+++ b/components/long-click.js
@@ -1,0 +1,99 @@
+// node_modules
+import { useEffect, useRef } from "react";
+
+/**
+ * Default amount of time in milliseconds to wait before considering a mouse click a long press.
+ */
+const DEFAULT_DELAY = 250;
+
+/**
+ * Hook to detect short and long clicks in an element. Pass it the ID of the element to attach the
+ * event listeners to. Pass it the regular- and long-click callbacks. It calls the regular-click
+ * callback when the user short-clicks the element, and the long-click callback when the user
+ * long-clicks the element.
+ *
+ * It returns an object with a function to call to trigger the mouse-up action, if some other way
+ * of triggering the mouse-up action is needed. As an example, a checkbox needs to handle clicks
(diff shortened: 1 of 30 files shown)


dougli1sqrd (approved):

forresttanaka:
@dougli1sqrd A _lot_ has changed since you reviewed this the first time. Things like the All and None buttons above each facet are the same, as is the Clear All button. But the ANY and NOT buttons are gone. We now have long clicks produce a negative facet term. The key file with this change is long-click.js which isn’t a big file but was complex to get working.

components/long-click.js
```diff
@@ -0,0 +1,99 @@
...
+ */
+export default function useLongClick(
+  elementId,
+  onClick,
+  onLongClick,
+  delay = DEFAULT_DELAY
+) {
+  // The "UNSET" timer value indicates that the timer isn't active. That's as opposed to the null
+  // value, which indicates that the timer has expired
+  const timer = useRef("UNSET");
```
dougli1sqrd, line 36: in typescript land I think we could use an enum or something to express this. Such is the js life though for now 🤷

components/long-click.js
```diff
@@ -0,0 +1,99 @@
...
+  function onMouseUp() {
+    if (timer.current) {
+      clearTimeout(timer.current);
+      onClick();
+    } else {
+      // The mouse-up event happened after the click timer expired, so the user long-pressed the
+      // checkbox. Call the onLongChange callback if it exists.
+      onLongClick();
+    }
+    timer.current = "UNSET";
```
dougli1sqrd, line 62: Maybe we should turn this into a constant in this file?

components/long-click.js
```diff
@@ -0,0 +1,99 @@
...
+   * released the mouse button before the click timer expired, call the `onClick` callback.
+   * Otherwise, the user long-pressed the checkbox, so call the `onLongClick` callback if it exists.
+   */
+  function onMouseUp() {
+    if (timer.current) {
+      clearTimeout(timer.current);
+      onClick();
+    } else {
+      // The mouse-up event happened after the click timer expired, so the user long-pressed the
+      // checkbox. Call the onLongChange callback if it exists.
```
dougli1sqrd, line 59: In this comment, do you mean `onLongClick`?

components/long-click.js
```diff
@@ -0,0 +1,99 @@
...
+    event.preventDefault();
+    if (event.detail === 0) {
+      onMouseUp();
+    }
+  }
+
+  useEffect(() => {
+    // Install the event listeners for the checkbox. We don't need the "mouseup" event because
+    // "click" fires after "mouseup," getting handled by the checkbox `onClick` handler.
+    const el = document.getElementById(elementId);
```
dougli1sqrd, line 82: Is there any possibility of running into a server side rendering issue here? `document` doesn't exist yet when rendered server right? Is it possible to attempt to run this on the server side?

Also out of curiosity, instead of using an elementId, is this where one might use `useRef`? If so, what's the difference?

components/long-click.js
```diff
@@ -0,0 +1,99 @@
...
+   * Otherwise, the user long-pressed the checkbox, so call the `onLongClick` callback if it exists.
+   */
+  function onMouseUp() {
+    if (timer.current) {
+      clearTimeout(timer.current);
+      onClick();
+    } else {
+      // The mouse-up event happened after the click timer expired, so the user long-pressed the
+      // checkbox. Call the onLongChange callback if it exists.
+      onLongClick();
```
dougli1sqrd, line 60: Also mostly as a curiosity, I wonder how this would feel to use if the long click handler happened automatically before the mouse was released? Like, if you held for the appropriate time delay, then the event fires because you've held, and then release doesn't do anything afterwards?

I was also curious about some visual cue that the long press timer was waiting. Like, if the delay was long (it's not by default so maybe this isn't an issue) I could imagine as a user not knowing how long I need to press. Having a visual that indicates when the long press is finished or ready could be helpful in some circumstances? Doesn't seem required here necessarily.

dougli1sqrd (changes requested):

components/long-click.js
forresttanaka: Yes.

components/long-click.js
forresttanaka: Yes.

components/long-click.js
forresttanaka: This code is inside a useEffect which hides the callback code from the server — the server can see this code, but can never execute it. It’s possible this should be a useRef. I’ll look at that.

dougli1sqrd (approved):

PR #322, opened 2023-10-10
IGVF-1174-getObject-refactor

dougli1sqrd:
* `FetchRequest.getObject` now returns `Result`. If the server response is not 2xx then it builds an error object and returns an `Err` Result. If fetch throws an error due to a network error, then the network error object is wrapped in Err and returned. After a successful fetch the object is wrapped in Ok and returned.
* `FetchRequest.getMultipleObjects` and `FetchRequest.getMultipleObjectsBulk` were also updated since they use `getObject`.
* `FetchRequest.getCollection` was updated since it uses `getObject`.
* All downstream callers of any of these methods were updated.
    * Generally the update simply calls `union()` on the returned result as that simply returns whatever underlying data (error or otherwise) is wrapped by the Result, which is what consumers were expecting before the Result change.
* Some tests were added, and any failing tests were fixed or corrected.

@forresttanaka There is still a present failing test that I'm unsure how to handle. The way the error handling works and in the tests is a little confusing, and in some of the tests I made Decisions about how the test should go. But please, if that wasn't the right call, let me know.


dougli1sqrd:
Also I forgot to add a file that I renamed / converted to ts, but I'm struggling with git, so that'll be incoming in due time

lib/general.ts
```diff
@@ -35,6 +35,10 @@ export function urlWithoutParams(url: string): string {
   return url.split("?")[0];
 }
 
+export function itemId(item: string | { "@id": string }): string {
+  return typeof item === "string" ? item : item["@id"];
+}
```
forresttanaka, line 40: JSDoc

forresttanaka (changes requested):
Looks great! I’m sure I’ll need help with any new code I write that uses these. Right now I’m least clear about when and why we would use the `Ok` and `Err` variants of `Result` instead of just `Result`?

Looks like common-requests.js needs reformatting. Might have other things too since the tests failed at the first step.

forresttanaka (approved):
Looking forward to getting this in. I don’t think there’s a QA step to this one either.

PR #323, opened 2023-10-12
IGVF-833-prediction-set-page

jenjou:


components/search/list-renderer/prediction-set.js
```diff
@@ -0,0 +1,43 @@
+// node_modules
+import PropTypes from "prop-types";
+// components/search/list-renderer
+import AlternateAccessions from "../../alternate-accessions";
```
forresttanaka, line 4: You actually get `AlternateAccessions` from `components`, not `components/search/list-renderer`, so it should really go in its own section under a `// components` comment.

forresttanaka (dismissed):
Cool! I have a very optional comment that’s safe to ignore if you choose to.

components/search/list-renderer/prediction-set.js
jenjou: Thanks. I copied that from auxiliary set, so it looks like the same issue is in that file too. I will fix it here, though.

forresttanaka (approved):

PR #324, opened 2023-10-13
IGVF-1178 Add GREGoR collection logo

forresttanaka:
I also fixed an unrelated crash with file-set tables. I changed the link that the logos point at to make more sense, as the collections aren’t indexed for site-wide search.


components/collections.js
```diff
@@ -51,7 +52,7 @@ export default function Collections({ collections = null }) {
           return (
             <Link
               key={collection}
-              href={`/site-search/?term=${collection}`}
+              href={`/search/?type=${itemType}&collections=${collection}&status=released`}
```
mingjiecn, line 55: Do we need status here since the user with certain access may also want to see data that is not released?

mingjiecn (dismissed):
Just have a question regarding using the status for search. Other than that it looks good.

forresttanaka:
@mingjiecn It’s possible. I based this on encode where nearly all searches include status=released. So we’ll just have to see if wranglers want this or not for igvf-ui.

PR #336, opened 2023-10-24
IGVF-1217-fix-alignment

dougli1sqrd:


dougli1sqrd:
This should be the only instance, I searched for
```
`\)[\n\s]*\.union
```
And that was the last one

forresttanaka (approved):

PR #338, opened 2023-10-24
IGVF-1082-authentication-typescript

dougli1sqrd:
~~I wasn't quite sure what to do with the auth0 functions, but also it looks like those are the only places where the auth0 functions are used. So let me know if using `any` could be better here, but I thought it was okay in this small instance since we're interfacing here with another package.~~

Nevermind, the linter is very mad at me

### DIFF
--- a/lib/collection-titles.js
+++ b/lib/collection-titles.js
@@ -8,5 +8,7 @@ import FetchRequest from "./fetch-request";
  */
 export default async function getCollectionTitles(dataProviderUrl) {
   const request = new FetchRequest();
-  return (await request.getObjectByUrl(`${dataProviderUrl}/collection-titles/`)).union();
+  return (
+    await request.getObjectByUrl(`${dataProviderUrl}/collection-titles/`)
+  ).union();
 }
--- a/lib/collection-titles.js
+++ b/lib/collection-titles.js
@@ -10,5 +10,5 @@ export default async function getCollectionTitles(dataProviderUrl) {
   const request = new FetchRequest();
   return (
     await request.getObjectByUrl(`${dataProviderUrl}/collection-titles/`)
-  ).union();
+  ).optional();
 }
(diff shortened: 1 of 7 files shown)


forresttanaka:
Tests fail.

lib/profiles.js
```diff
@@ -5,9 +5,9 @@ import FetchRequest from "./fetch-request";
  * Loads the schemas for all object types, with each key of the object being the @type for each
  * schema.
  * @param {object} session Authentication session object
- * @returns {Promise<object>} Promise that resolves to the /profiles object
+ * @returns {Promise<DataProviderObject | ErrorObject>} Promise that resolves to the /profiles object
  */
 export default async function getProfiles(dataProviderUrl) {
   const request = new FetchRequest();
-  return request.getObjectByUrl(`${dataProviderUrl}/profiles`, null);
+  return (await request.getObjectByUrl(`${dataProviderUrl}/profiles`)).union();
```
forresttanaka, line 12: Should this be `optional()` since the previous default is null, and session-context.js doesn’t really do anything with an error.

lib/collection-titles.js
```diff
@@ -8,5 +8,7 @@ import FetchRequest from "./fetch-request";
  */
 export default async function getCollectionTitles(dataProviderUrl) {
   const request = new FetchRequest();
-  return request.getObjectByUrl(`${dataProviderUrl}/collection-titles/`, null);
+  return (
+    await request.getObjectByUrl(`${dataProviderUrl}/collection-titles/`)
+  ).union();
```
forresttanaka, line 13: Same question here as for getProfiles.

forresttanaka (changes requested):
I’m setting to “Request changes,” but my feedback is more like questions. If you confirm what you have is correct, then I’ll approve.

lib/profiles.js
dougli1sqrd: oh yes, you're right

lib/collection-titles.js
dougli1sqrd: yep

forresttanaka (approved):

forresttanaka:
Hey, I’m starting to get it (optional(), union(), etc.)!

PR #353, opened 2023-11-02
IGVF-1233 Display embedded properties in report cells

forresttanaka:
I changed `DownloadTSV` because it sometimes ran on the server, where `window` isn’t defined, leading to a crash.

I converted report.js to Typescript, which turned out far more complicated than I anticipated. The main issue is using syntax like `object[index]` seemed to need to split the object interface into static and index signature interfaces, and then combine them into object. Let me know if you have a better idea, because this seems unbelievably clumsy.


globals.d.ts
```diff
@@ -52,6 +52,137 @@ export interface DatabaseObject {
   [key: string]: unknown;
 }
 
+/**
+ * Search result object; similar to `DatabaseObject` but without some properties.
+ */
+export type SearchResultsObject =
```
dougli1sqrd, line 58: I'd say something similar here as for Profiles below, I think

globals.d.ts
```diff
@@ -52,6 +52,137 @@ export interface DatabaseObject {
...
+    [key: string]: {
+      [key: string]: unknown;
+    };
+  };
+  description?: string;
+  exact_searchable_fields?: string[];
+  fuzzy_searchable_fields?: string[];
+  identifyingProperties?: string[];
+  mixinProperties: object[];
+  properties: SchemaProperties;
```
dougli1sqrd, line 104: Maybe I'm missing something, but I think this might be off? When I'm looking at the `properties` field of a schema, I'm seeing:
```
prop_name: {
  "title": string,
  "type": string,
   "description": string,
   // etc
}
```
So I think for `properties` you'd want something closer to:
```
// ...
properties:   [key: string]: SchemaProperties;
// ...
```

globals.d.ts
```diff
@@ -52,6 +52,137 @@ export interface DatabaseObject {
...
+}
+
+export interface SchemaPropertiesGenerics {
+  [key: string]: unknown;
+}
+
+/**
+ * Describes schema objects.
+ */
+export interface Schema {
```
dougli1sqrd, line 88: It's also occurring to me that these are all JSON-schema and JSON-LD. Especially for `properties` field, I wonder if there are typescript types for JSON-schema already made?

globals.d.ts
```diff
@@ -52,6 +52,137 @@ export interface DatabaseObject {
...
+  required: string[];
+  title: string;
+  type: string;
+}
+
+/**
+ * Describes the /profiles endpoint object, which is an object with `@type`s as its keys and
+ * the corresponding schemas for each type as the values.
+ */
+export type Profiles = ProfilesProps | ProfilesGeneric;
```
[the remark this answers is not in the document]
dougli1sqrd: I double checked that you can do this: https://www.typescriptlang.org/play?#code/C4TwDgpgBAygxgCwgWwIYAUBOB7SngCWEAzlALyyIoY56Ela6kA+lSajdRxA4hAHYRMBOMQDcAKAkF+wIQDNUcaPHY1cQ+sU6kA3hKhRCwADYQAXFGLBh-AOaTDoSJeu2HEgL5SZczIuU2ak5Nbj5BYVEofUMAbQBrCBBXGxk7AF1LAFd+eP5sAHd+SW8JOGx+YmwzADoTbDsACgAiJBN65oBKSWlZBSVoAFkQEPxuaIMjAlMLK1T7RyNwWbc0xYSklPdMqBy8wuKvKXkcuEIK3eIITkawWkth0a1OicNyyuqIOobGgAMAFWmZksABJdHdcDVjGZPL9upN3lVavUmgDlqDwbQoctYfC3hUkV8UX8ABKoUhggDyACMAFYQM41RIgYi3WidOoCOzABBQAC0UAATJ4oNgeUIoMziHCSlIslcbjEpjNLM0AIJQHiYVD8AAmUEBM2aABpJs5Zs0AKIAD1QyDAZhNk3k2GwqupqEwTs88KAA

globals.d.ts
```diff
@@ -52,6 +52,137 @@ export interface DatabaseObject {
...
+ * The `properties` object of a schema.
+ */
+export type SchemaProperties = SchemaPropertiesProps | SchemaPropertiesGenerics;
+
+export interface SchemaPropertiesProps {
+  title: string;
+  type: string;
+}
+
+export interface SchemaPropertiesGenerics {
```
dougli1sqrd, line 81: If we we used known json-schema types, or copied them, I don't think we'd have to have a generic key: unknown since we'd know the types of properties and what values they take and we'd have to cast less I think. But that might be overkill as well.

dougli1sqrd (changes requested):
Left some comments mostly in global.d.ts.

I think the other more general thing I'll say here is that I although typescript likes and is fine with optional properties, it definitely goes against my intuition on how I would think about types and data. To me if a field needs to be optional, or "nullable", the existence of the property should not be optional, but the value of the property should be: `T | null` where `T` is whatever type the optional property would take.

It's the same thing either way I guess, except that an interface that says you _must_ have property `foo` and its value is `T | null` is slightly more restrictive than `foo?: T` since `{}` and `{foo: intanceOfT }` both type check, but `foo: T | null` would not type check `{}`. And I think I prefer the slight restrictiveness because the more restrictive your types are the more assumptions you can make about code that uses that restrictive type.

To handle the situation where your data coming in may not, in fact, have `foo` defined in the object, the way I'd think about is that there would be a parsing step, that takes the raw data and attempts to bring it into your typed "walled garden" so to speak. So in the above foo example, if the backend gave us `{}`, but we should except that as a null `foo`, the parse step would take the "untyped" data and convert it: taking `{}` => `{ foo: null }` thus type checking `foo: T | null`.

I think this is largely a javascript/typescript cultural divide though, so I'm definitely not asking to change anything with optional properties in interfaces, just offering an alternative point of view from other languages

lib/report.ts
```diff
@@ -0,0 +1,250 @@
...
+ * @param {Profiles|null} profiles All schemas for all types keyed by their `@type`
+ * @returns {Schema[]} Schemas for the given `reportTypes`
+ */
+export function getSchemasForReportTypes(
+  reportTypes: string[],
+  profiles: Profiles | null
+): Schema[] {
+  if (profiles) {
+    return reportTypes.reduce((schemaAcc: Schema[], reportType) => {
+      const schema = (profiles as ProfilesGeneric)[reportType];
```
dougli1sqrd, line 43: I think now that `Profiles` is both `ProfilesGeneric` and `ProfilesProps`, you shouldn't have to cast here?

dougli1sqrd: (or below)

dougli1sqrd (changes requested):

lib/report.ts
forresttanaka: Yeah, works great without the cast.

lib/report.ts
```diff
@@ -0,0 +1,256 @@
...
+ * null gets returned.
+ * @param {string[]} reportTypes All `type` in the query string of the report
+ * @param {Profiles|null} profiles All schemas for all types keyed by their `@type`
+ * @returns {Schema[]} Schemas for the given `reportTypes`
+ */
+export function getSchemasForReportTypes(
+  reportTypes: string[],
+  profiles: Profiles | null
+): Schema[] {
+  if (profiles) {
```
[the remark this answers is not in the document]
forresttanaka: I avoid early return more than most, probably because of my early training where early return was frowned upon, as it’s kind of a form of goto, which was also frowned upon, and still is. So basically, if the situation feels convenient to avoid early return, I avoid it. Plus if you want to console.log data at the end of a function, you have to search through the function for early returns that also need a console.log, which I find a little painful. I’d like to avoid all early returns, but I know I could never get away with it these days.

dougli1sqrd (approved):

PR #367, opened 2023-11-16
IGVF-1079-construct-library-summary

dougli1sqrd:
adding construct library sets to multiplexed samples, prediction sets, and measurement sets

### DIFF
--- a/pages/multiplexed-samples/[uuid].js
+++ b/pages/multiplexed-samples/[uuid].js
@@ -1,5 +1,6 @@
 // node_modules
 import { TableCellsIcon } from "@heroicons/react/20/solid";
+import { Fragment } from "react";
 import PropTypes from "prop-types";
 // components
 import AlternateAccessions from "../../components/alternate-accessions";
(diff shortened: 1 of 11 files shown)


pages/multiplexed-samples/[uuid].js
```diff
@@ -65,6 +75,38 @@ export default function MultiplexedSample({
...
                   </>
                 )}
+                {constructLibrarySetCollapser.displayedData.length > 0 && (
+                  <>
+                    <DataItemLabel>Construct Library Sets</DataItemLabel>
+                    <DataItemValue>
+                      <SeparatedList>
+                        {constructLibrarySetCollapser.displayedData.map(
+                          (acc, i) => (
+                            <Fragment key={i}>
```
Comment (forresttanaka) [inline pages/multiplexed-samples/[uuid].js:85]: Use `acc["@id"]` as the key.

forresttanaka (changes requested):
A couple small changes. Also, I feel like the request for samples should go into common-requests.js. If we end up needing more than the `construct_library_sets` property, we’d have to change that code in two places as it stands.

forresttanaka (approved):

PR #370, opened 2023-11-20
IGVF-947 Add more error logging

forresttanaka:
Lots of file changes, but the vast majority happened because I changed a default export of the errors.js file to a named export, so many pages had to change their import correspondingly.

### DIFF
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -6,12 +6,23 @@
 import { HTTP_STATUS_CODE } from "./fetch-request";
 
 /**
+ * Log the error to the console with a timestamp.
+ * @param {string|number} statusCode Error status code, e.g. 403
+ * @param {string} title Narrative string corresponding to error code
+ * @returns {void}
+ */
+export function logError(statusCode, title) {
+  const date = new Date().toISOString();
+  console.log(`ERROR [${date}] ${statusCode}:${title}`);
+}
+
+/**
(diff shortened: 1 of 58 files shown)


lib/errors.js
```diff
@@ -5,13 +5,24 @@
...
 
+/**
+ * Log the error to the console with a timestamp.
+ * @param {string|number} statusCode Error status code, e.g. 403
+ * @param {string} title Narrative string corresponding to error code
+ * @returns {void}
+ */
+export function logError(statusCode, title) {
+  const date = new Date().toISOString();
+  console.log(`ERROR [${date}] ${statusCode}:${title}`);
```
dougli1sqrd, line 16: Should this be `console.error` instead of `log`? `Error` will print to standard error instead of standard out. Is that a distinction on web pages?

dougli1sqrd (dismissed):
Just a question about std error vs out. If we should do `console.error` then I'd say this is "Request Changes", but otherwise it looks fine

lib/errors.js
forresttanaka: AFAIK there’s no distinction in the server logs. Browsers can display them in different colors, but I don’t think anything beyond that.

forresttanaka: I’ll leave them as is. We’ve had other error messages that use console.log, so if we decide they should use `console.error`, then we can convert all of them at once.

PR #375, opened 2023-12-05
IGVF-1182-sample-page-UI

khzlin:


components/common-data-items.js
```diff
@@ -115,7 +115,12 @@ DonorDataItems.commonProperties = [
 /**
  * Display data items common to all sample-derived objects.
  */
-export function SampleDataItems({ item, sources = null, children }) {
+export function SampleDataItems({
+  item,
+  sources = null,
+  children,
+  sortedFractions,
```
forresttanaka, line 122: Always put `children` last because it gets treated differently from other props.

components/common-data-items.js
```diff
@@ -164,18 +169,28 @@ export function SampleDataItems({ item, sources = null, children }) {
           </DataItemValue>
         </>
       )}
-      {item.sorted_fraction && (
+      {item.sorted_from && (
         <>
-          <DataItemLabel>Sorted Fraction</DataItemLabel>
+          <DataItemLabel>Sorted From Sample(s)</DataItemLabel>
```
forresttanaka, line 174: It looks like this always gets sorted from a single sample. Should this have the “(s)”?

components/common-data-items.js
```diff
@@ -164,18 +169,28 @@ export function SampleDataItems({ item, sources = null, children }) {
...
-              {item.sorted_fraction.accession}
+            <Link href={item.sorted_from["@id"]}>
+              {item.sorted_from.accession}
             </Link>
-            {item.sorted_fraction_detail ? (
-              <> {item.sorted_fraction_detail}</>
-            ) : (
-              ""
-            )}
+            {item.sorted_from_detail ? <> {item.sorted_from_detail}</> : ""}
```
forresttanaka, line 179: This is correct, but we usually do `{items.sorted_from_detail && <> {item.sorted_from_detail}</>}` as a little shortcut, and as kind of a standard for conditional text.

components/common-data-items.js
```diff
@@ -164,18 +169,28 @@ export function SampleDataItems({ item, sources = null, children }) {
...
+          </DataItemValue>
+        </>
+      )}
+      {sortedFractions?.length > 0 && (
+        <>
+          <DataItemLabel>Sorted Fractions of Sample(s)</DataItemLabel>
```
forresttanaka, line 185: Another option, if you’re up for it, is to use the singular “Sample” if `sortedFractions.length === 1`, but “Samples” otherwise.

components/common-data-items.js
```diff
@@ -352,7 +376,7 @@ export function BiosampleDataItems({
       </>
       {pooledFrom?.length > 0 && (
         <>
-          <DataItemLabel>Biosample(s) Pooled From</DataItemLabel>
+          <DataItemLabel>Pooled From Sample(s)</DataItemLabel>
```
forresttanaka, line 379: Similar thing for plurals as above. I know it was done with “(s)” before, as well as below.

components/common-data-items.js
```diff
@@ -364,9 +388,37 @@ export function BiosampleDataItems({
...
+                </Link>
+              ))}
+            </SeparatedList>
+          </DataItemValue>
+        </>
+      )}
       {partOf && (
         <>
-          <DataItemLabel>Part of Biosample</DataItemLabel>
+          <DataItemLabel>Part of Sample(s)</DataItemLabel>
```
forresttanaka, line 421: Similar to above, I’m suprised this can be plural. I might just not understand.

forresttanaka (changes requested):
Whoa, that looked like a lot of work. None of my feedback is big, and some were questions because I wasn’t sure.

components/common-data-items.js
khzlin: done

components/common-data-items.js
khzlin: done

components/common-data-items.js
khzlin: done

components/common-data-items.js
khzlin: confirmed this one is single. fixed.

forresttanaka (approved):

components/common-data-items.js
```diff
@@ -183,7 +183,7 @@ export function SampleDataItems({
       {sortedFractions?.length > 0 && (
         <>
           <DataItemLabel>
-            Sorted Fractions of Sample{sortedFractions.length === 1 ? "" : "s"}
+            Sorted Fraction of Sample{sortedFractions.length === 1 ? "" : "s"}
```
ian-whaling, line 186: I think this should be "Sorted Fraction Sample{sortedFractions.length === 1 ? "" : "s"}"

khzlin: @ian-whaling your comment gave me the option of either so I picked one `I think the title should be renamed to “Sorted Fraction Sample(s)” or “Sorted Fraction(s) of Sample”.`

ian-whaling: or "Sorted Fraction{sortedFractions.length === 1 ? "" : "s"} of Sample"

khzlin: talked to Ian offline and we both agreed to `Sorted Fraction{sortedFractions.length === 1 ? "" : "s"} of Sample`

forresttanaka (approved):

PR #378, opened 2023-12-11
IGVF-1313-image-file-page

khzlin:


forresttanaka (changes requested):
You need to change the `files.test.ts` Jest test to get tests to pass, because it fails full coverage right now. You probably need to add a test for the case where `file["@type"][0]` contains `ImageFile`.

forresttanaka (approved):

PR #380, opened 2023-12-11
IGVF-1310 Add formatted schema displays

forresttanaka:
This is kind of a big PR; no two ways about it. The main concern is pages/profiles/[profile].js, but it also contains a new tabs module along with documentation in a markdown file, and a small json-panel.js module. It also adds a new Cypress test, as well as new Tailwind CSS class definitions.

Going forward, I’m not expecting highly detailed code reviews, and I don’t provide them either. Just whatever you come across that seems wrong or not optimal will have a lot of utility. I tend to focus on code/operational duplication. But we’re all busy so we can only do what we can in the small amount of time we have.

I’m writing up QA notes for the ticket that I’ll post shortly. You can try the demo if you like.


pages/profiles/[profile].js
```diff
@@ -1,21 +1,470 @@
...
+      // the user holds down the option key.
+      if (isOptionKey) {
+        setOpenJsonIds(Object.keys(dependencies));
+      } else {
+        setOpenJsonIds([...openJsonIds, propertyId]);
+      }
+    }
+  }
+
+  if (dependencies) {
```
Comment (mingjiecn) [inline pages/profiles/[profile].js:155]: somehow there is empty dependentSchemas in some of the schemas. So we should also check if the dependentSchemas is empty. Check the biomarker has empty dependentSchemas: https://igvf-ui-igvf-1310-formatted-schemas.demo.igvf.org/profiles/biomarker/

mingjiecn (changes requested):
Please check my single comment. Thank you!

pages/profiles/[profile].js
Comment (forresttanaka) [reply pages/profiles/[profile].js]: Did not even realize `dependentSchemas` wasn’t consistently defined.

mingjiecn (dismissed):

PR #388, opened 2024-01-05
IGVF-1085 Biomarker/Breadcrumbs to Typescript

forresttanaka:
I still don’t have the handle on Typescript that Eric had, so we’ll just all need to learn this together, and be happy to update things later as we learn more.

I’m still planning on limiting igvf-ui’s Typescript conversion to the lib/ directory so that wranglers will not likely deal with Typescript.


lib/breadcrumbs.ts
```diff
@@ -126,8 +158,14 @@ function buildSearchResultBreadcrumbs(data) {
...
 
-function buildSchemaBreadcrumbs(schema, profile) {
-  if (!profile) {
+/**
+ * Generate the breadcrumb data for a schema page.
+ * @param schema {DatabaseObject | null} The schema object, or null for the schema index page
+ * @param title {string} The data type associated with a schema, e.g. "Biomarker"
+ * @returns {Breadcrumb[]} Breadcrumb data for the given schema
+ */
+function buildSchemaBreadcrumbs(schema: Schema, title: string): Breadcrumb[] {
```
mingjiecn, line 167: if the schema can be null, should be `schema?: Schema`

lib/__tests__/breadcrumbs.test.ts
```diff
@@ -27,31 +30,37 @@ describe("Test breadcrumb composition and rendering functionality", () => {
...
+      "@context": "/terms/",
       name: "j-michael-cherry",
       "@id": "/labs/j-michael-cherry/",
       "@type": ["Lab", "Item"],
       title: "J. Michael Cherry, Stanford",
+      status: "current",
+      uuid: "d91b048e-2d8a-4562-893d-93f0e68397c0",
     };
 
-    let breadcrumbs = await buildBreadcrumbs(labItemData, labItemData.title);
```
mingjiecn, line 37: Why `labItemData.title` is not working anymore in typescript?

mingjiecn (changes requested):
Still learning TS. I have a couple questions in the comment. Thank you so much for your help!

lib/breadcrumbs.ts
forresttanaka: Oh! That’s a comment mistake. `buildBreadcrumbs` only gets called when `data` is known to be a good object.

lib/__tests__/breadcrumbs.test.ts
forresttanaka: Yeah, I didn’t use the right solution there. I should have typecast the argument, which I now do. Really weird that Typescript can’t see that the `title` property is a string though. I assign a string to `title`, and Typescript thinks that’s a mystery?

mingjiecn: I guess Typescript is not that clever or just very strict...

mingjiecn (approved):

PR #395, opened 2024-01-18
IGVF-1393-display-assay-details

khzlin:


pages/measurement-sets/[id].js
```diff
@@ -279,18 +278,18 @@ export default function MeasurementSet({
...
-                      <DataItemValueUrl>
-                        <a
-                          href={measurementSet.protocol}
-                          target="_blank"
-                          rel="noopener noreferrer"
-                        >
-                          {measurementSet.protocol}
-                        </a>
-                      </DataItemValueUrl>
+                      <DataItemLabel>Protocol(s)</DataItemLabel>
```
Comment (forresttanaka) [inline pages/measurement-sets/[id].js:283]: I’d like to avoid using the “(s)” suffix in IGVF and instead either always use plural even if only one exists, or to make the pluralization conditional based on the count. I think always using “Protocols” here works even if only one exists. The conditional case works better if you have something like “You have {x} protocol.”

forresttanaka (changes requested):
Looks great! I just have one style preference comment.

pages/measurement-sets/[id].js
Comment (khzlin) [reply pages/measurement-sets/[id].js]: done

forresttanaka (approved):

PR #396, opened 2024-01-19
IGVF-1319 Display sample terms in report cells

forresttanaka:


mingjiecn (changes requested):
I think we also need to add this function to the parent types Sample and Biosample. Other than that it looks good.

forresttanaka:
> I think we also need to add this function to the parent types Sample and Biosample…

Oh! I forgot that was possible.

PR #402, opened 2024-01-24
IGVF-1305 Redesign home page

forresttanaka:
When displaying the chart while using a dev build, you see a console warning, “Failed prop type: Converting circular structure to JSON.” As described in [this thread](https://github.com/facebook/prop-types/issues/383), this seems like a bug in PropTypes. PropTypes don’t run in production builds, so that causes no problems there.


components/navigation.js
```diff
@@ -406,6 +406,46 @@ NavigationHrefItem.propTypes = {
...
+  navigationClick,
+  isChildItem = false,
+  isNarrowNav = false,
+  children,
+}) {
+  const cssClasses = navigationClasses(isNarrowNav, isChildItem);
+
+  return (
+    <li>
+      <a
```
mingjiecn, line 424: Since it is an external site, should it open in a new tab instead of the original tab?

forresttanaka: I usually think so, but the igvf.org site links to our page without opening a new tab, so I felt maybe it makes sense to return to that page without opening a new tab.

forresttanaka: Oh wait, you’re talking about the catalog. Yeah, I thought about that, but I really wasn’t sure.

components/chart-file-set.js
```diff
@@ -0,0 +1,328 @@
...
+  const { darkMode } = useContext(GlobalContext);
+  const legendTextColor = darkMode.enabled ? "#ffffff" : "#000000";
+
+  // Filter the file sets by the selected month, then convert to a form the Nivo bar chart can use.
+  const selectedFileSets = filterFileSetsByMonth(fileSets, selectedMonth);
+  const { fileSetData, maxCount } = convertDataToNivo(selectedFileSets);
+
+  // The legend should only appear while the user is logged in.
+  const { isAuthenticated } = useAuth0();
+  const legendProps = isAuthenticated
```
mingjiecn, line 245: Why legend is hidden if not logged in? I don't feel that this is sensitive data.

forresttanaka: The legend shows the color differences between released and non-released file sets in the chart. But while logged out you only see released file sets, so Meenakshi felt it didn’t make sense to show this legend when you’d only ever see one color in the chart.

mingjiecn (dismissed):
Looks good to me. I do have a couple questions. Please check my comment. I also found some strange behavior in Safari. I can login using Safari no problem. But it will log me out automatically within a second. Not sure if you have this problem/

forresttanaka:
Thanks! On Safari, that’s a known issue when you have “Prevent cross-site tracking” enabled in Privacy settings. I have a solution to this, but I can only test it in a demo on weekends. That’s why I ran into an issue of my demos getting automatically deleted as soon as they came up.

forresttanaka:
Up for code review for the second time with an entire new chart added. You’ll see I moved the Tailwind CSS classes for light and dark mode text color from inside the `<App>` component to the `<body>` element so that it could also cover the dropdown portal root as well as future portal roots. The demo is up and updated.

mingjiecn (changes requested):
right now for the Data Set Releases Chart, the Y-axis show some non integer values, for example 2.5. Does it mean 2.5 or 2.5K FileSets? I think we should not show labels that is not a integer.

forresttanaka:
@mingjiecn Yeah, I know about that, but this goes away once we get more than a handful of released data sets, so given the difficulty of adding a custom Y-tick generator (I have an X-tick generator for the other chart because that issue could linger for a while), I felt it was OK not to fix this.

Am I being too cavalier?

mingjiecn:
Sounds reasonable. I will go ahead to approve it.

mingjiecn (dismissed):

forresttanaka:
@mingjiecn Thanks! I’m completely wrong though. After you said that, I looked into it more, and for this chart’s Y axis, the fix is way simpler than the more complicated X axis of the bar chart. In fact, it’s just the addition of one more integer attribute in the Nivo config. So I’ll make that change. Since the fix is so simple, I’ll just make it and go to QA.

Thanks for making me look at that again! I had figured it would be as complicated as the X-axis case for the bar chart, so it wasn’t worth worrying about.

PR #404, opened 2024-01-25
IGVF-1330 Fix multiple-type display on search pages

forresttanaka:
I changed a lot of the `doc_count` properties to not required. During testing of this branch, I found certain cases where searches returned terms that didn’t include `doc_count`, which caused no run-time issue (simply displays a blank as desired) except the required `doc_count` propType would display a console warning that it didn’t exist, though only on development builds, not production ones.


forresttanaka:
Oh wait, Jest coverage failed. I’ll set Jira to Code Review Ready once I fix that.

mingjiecn (changes requested):
I am not sure if I understand the purpose of this ticket. Looks like ui will not display types info at all if there is no keyword `type` in search query? I think the user may still want to see what types of data is returned. You are checking the filters field in returned results for types info. But actually in the `facets` field, there is a `type` field, which should have all the types info. I feel that this is what you need if you want to know the type info for the returned data. 
If what I mentioned above is not what you need, then there are couple other issues. First you should still check the returned results in graph to make sure it is not empty before you display types. Second we need to figure out how to display the types when they are not exclusive, for example: Donor, HumanDonor. Should we display `Subtypes of Donor, HumanDonor`, or should we just display `Subtypes of Donor`

mingjiecn (approved):

PR #415, opened 2024-02-07
IGVF-1285 Fix window resize crash after viewing schema SVG page

forresttanaka:
Interesting issue that got past me during code review. The function instance passed when adding the resize event listener isn’t the same instance as the one getting removed. I use `useCallback` to make sure the same instance gets passed when adding and removing.

I also took this chance to remove the custom hook and integrate its code into the main component. This custom hook doesn’t get shared and likely never will, so we don’t need a custom hook.

### DIFF
--- a/cdk/infrastructure/config.py
+++ b/cdk/infrastructure/config.py
@@ -53,6 +53,7 @@ config: Dict[str, Any] = {
                 'desired_count': 1,
                 'max_capacity': 4,
             },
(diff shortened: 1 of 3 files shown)


forresttanaka:
@mingjiecn Sorry, I need to make a small change to this branch. I forgot `useCallback` needs a second parameter. Doesn’t seem to cause problems not to have it, but it’s technically incorrect.

mingjiecn (approved):

PR #439, opened 2024-03-12
IGVF-1499 Integrate pager into sortable grid component

forresttanaka:
While the multireport page uses `SortableGrid` and has a table count and pager, it doesn’t use the ones now built into `SortableGrid`. I’d like to add parent-controlled counts and paging to `SortableGrid` in a future ticket.

I also disabled or fixed sorting on some table columns, as I noticed they didn’t sort correctly. I also added a new parameter to the sorting callback in the columns definition.

Unrelated but also fixed is a crash in `sortColumnSpecs` when schema properties don’t have a title. They’re all supposed to, but at least some still appear missing.


pedrohr (approved):
looks good!! 💯

PR #443, opened 2024-03-15
IGVF-1384 Have path arrays collapse in report views

forresttanaka:
The main point of this ticket is implemented in `<PathArray>` which now uses the collapser mechanism. But this revealed that I had been using my own SortableGrid `display()` method incorrectly, calling the display component directly as if it were just a regular function.

To fix this involved fixing many other tables to use `display()` correctly by passing the `meta` property as a React property instead of a second parameter. The bulk of the file changes in this branch involve those changes. Without this change, adding a state to the report view cell display caused a crash.

The actual fix to this improper call of the display method is in `<DataGrid>`.

I also reversed the order of the parameters to the `useCollapseControl` custom hook to seem more natural.

I removed the ExternalResources component, as it displayed a property on donor pages that doesn’t exist.


components/sortable-grid.js
```diff
@@ -184,13 +184,11 @@ function HeaderCell({ cells, cellIndex, meta, children }) {
   // header cell title renderers. A child of type "object" indicates a React component, while
```
mingjiecn, line 184: I feel that you may need to update the comment here since you don't distinguish whether the child is object or not any more?

mingjiecn (dismissed):

PR #445, opened 2024-03-18
IGVF-1455-audit-page-polish

khzlin:


pages/audits.js
```diff
@@ -11,13 +12,34 @@ import FetchRequest from "../lib/fetch-request";
...
+    {
+      audit_level: "WARNING",
+      audit_description:
+        "Possibly inconsistent metadata. Data can be released with warnings",
+    },
+    {
+      audit_level: "INTERNAL_ACTION",
+      audit_description: "Metadata errors that require DACC staff to resolve",
+    },
+  ];
```
forresttanaka, line 42: Since `auditKeyColor` never changes, it’s best to put these outside of functions so that it only gets executed on page load. Putting it in the function means it gets executed every render.

forresttanaka (dismissed):
I have a couple items, but I’d call them optional. Of the two, the `auditKeyColor` one is more important. But it works; it’s just a tiny bit of an inefficiency, so change that only if you feel like it.

components/audit-table.js
```diff
@@ -1,46 +1,35 @@
...
-            );
-          })}
+        <div className="flex items-center justify-start gap-1">
+          <mapping.Icon className={`h-4 w-4 ${mapping.color}`} />
+          <div>{mapping.humanReadable}</div>
         </div>
       );
     },
-    isSortable: false,
+    isSortable: true,
```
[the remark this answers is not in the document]
khzlin: ok

pages/audits.js
khzlin: ok

forresttanaka (approved):

PR #449, opened 2024-03-25
IGVF-1353-update-schema-page

forresttanaka:
I had to change the name of the [profile].js to [...profile].js so that I could access the rest of the path after the schema collection name part of the path from `params.profile`.


pages/profiles/[...profile].js
```diff
@@ -571,18 +747,77 @@ Changelog.propTypes = {
...
+ */
+function SchemaTab({ id, tabTitle, tabUrl = "", children }) {
+  const { isCopied, initiateCopy } = useCopyAction(tabUrl);
+  const tooltipAttr = useTooltip(`copy-tab-url-${id}`);
+
+  return (
+    <div className="relative">
+      <TabTitle id={id}>{children}</TabTitle>
+      <TooltipRef tooltipAttr={tooltipAttr}>
+        <button
```
Comment (mingjiecn) [inline pages/profiles/[...profile].js:761]: Why not use CopyButton here?

mingjiecn (changes requested):

pages/profiles/[...profile].js
Comment (forresttanaka) [reply pages/profiles/[...profile].js:761]: I felt this copy button has to be as unobtrusive as possible, and even the new `sm` option on the copy button was too big and bold to use in these tabs where you normally see nothing unless you hover over the right spot (I don’t actually expect many submitters to use these links — mostly I expect wranglers to use them; I could be wrong).

So I followed the same principle I have for using the `form-elements/button`  module — use it if you can. Do something else if you can’t.

Comment (mingjiecn) [reply pages/profiles/[...profile].js:761]: If only wranglers use it then it should not be a problem. I just had a hard time to find those copy buttons.

mingjiecn (approved):

pages/profiles/[...profile].js
Comment (forresttanaka) [reply pages/profiles/[...profile].js:761]: Thanks. We’ll see what happens in QA.

PR #458, opened 2024-04-01
IGVF-1543-orf-ui-page

khzlin:


components/search/list-renderer/open-reading-frame.js
```diff
@@ -0,0 +1,42 @@
...
+  SearchListItemMain,
+  SearchListItemMeta,
+  SearchListItemQuality,
+  SearchListItemTitle,
+  SearchListItemType,
+  SearchListItemUniqueId,
+} from "./search-list-item";
+
+export default function OpenReadingFrame({ item: openReadingFrame }) {
+  const geneId = openReadingFrame.gene?.map((gene) => gene.geneid).join(", ");
```
forresttanaka, line 15: `gene` is required, so you can just to `openReadingFrame.gene.map`. Too bad it’s not called `genes`. If you don’t feel like removing the question mark, it won’t hurt — it’s just not necessary.

pages/open-reading-frames/[uuid].js
```diff
@@ -0,0 +1,141 @@
...
+import FetchRequest from "../../lib/fetch-request";
+import { isJsonFormat } from "../../lib/query-utils";
+
+function EnsemblLink({ ensemblid }) {
+  return (
+    <a href={`http://www.ensembl.org/Homo_sapiens/Gene/Summary?g=${ensemblid}`}>
+      {ensemblid}
+    </a>
+  );
+}
```
Comment (forresttanaka) [inline pages/open-reading-frames/[uuid].js:29]: It’d be nice to move this function to its own file so you can share this component between this and the genes page, and we don’t duplicate the code. I know it‘s not _quite_ duplicate, but you just have to pass the `Homo_sapiens` taxa for this one.

forresttanaka (changes requested):
Mostly I’d like to avoid duplicating code, so if you could make a new ensemble-link.js file and move the one from gene there, and then import it into the gene page and the orf page, that would be great.

components/search/list-renderer/open-reading-frame.js
khzlin: done

pages/open-reading-frames/[uuid].js
Comment (khzlin) [reply pages/open-reading-frames/[uuid].js]: done

forresttanaka (dismissed):
Awesome — thanks!

pages/open-reading-frames/[uuid].js
```diff
@@ -63,30 +70,38 @@ export default function OpenReadingFrame({ orf, isJson }) {
                   </DataItemValue>
                 </>
               )}
-              {orf.pct_identical_protein && (
+              {orf.pct_identical_protein >= 0 && (
```
Comment (forresttanaka) [inline pages/open-reading-frames/[uuid].js:73]: We have a `truthyOrZero` function in lib/general.js to do this, but I’m thinking your way here is easier to understand. So let’s leave this as is. I might convert other things from `truthyOrZero` to this over time.

forresttanaka (approved):

PR #460, opened 2024-04-02
IGVF-1500-fix-environment-treatment-display

khzlin:


pages/treatments/[uuid].js
```diff
@@ -167,9 +167,13 @@ export async function getServerSideProps({ params, req, query }) {
     const documents = treatment.documents
       ? await requestDocuments(treatment.documents, request)
       : [];
+    let treatmentId = "";
+    treatment.treatment_type === "environmental"
+      ? (treatmentId = treatment.summary)
+      : (treatmentId = treatment.treatment_term_id);
```
Comment (forresttanaka) [inline pages/treatments/[uuid].js:173]: Conditional expressions always return a value, which you can use here:
```
const treatmentId = treatment.treatment_type === "environmental"
    ? treatment.summary
    : treatment_term_id;
```
That way you don’t have assignments within an expression. Generally, there should be only one assignment in an assignment statement, to a single variable on the left side of an equals sign.

forresttanaka (changes requested):

pages/treatments/[uuid].js
Comment (khzlin) [reply pages/treatments/[uuid].js]: i see. i'll fix it.

forresttanaka (approved):

PR #488, opened 2024-05-02
IGVF-1617 Collapsible lists and report links

forresttanaka:
I deleted the renderers for the `external_resources` property, as this property was removed from the human donor and rodent donor schemas in February 2023.

I’ve started to remove the conditional pluralization from data panel labels because I’d rather the label represent what that property _is_ instead of what that property _has_. I’ll still use pluralizations in other places, such as with the item count at the top of tables.


forresttanaka:
Hang on; I didn’t realize a Cypress test failed. I think it’s a fake failure. I’ll set the ticket to Code Review Ready once Cypress passes.

mingjiecn (dismissed):

PR #492, opened 2024-05-06
IGVF-1611 Hide internal action audit facet when not logged in

forresttanaka:


components/facets/custom-facets/audit-internal-action-terms.js
```diff
@@ -0,0 +1,36 @@
+// node_modules
+import { useAuth0 } from "@auth0/auth0-react";
+import PropTypes from "prop-types";
+// components/facets/custom-facets
+import StandardTerms from "./standard-terms";
+
+/**
+ * Display facet terms for internal action audit events but only for authenticated users.
+ */
+export default function InternalActionAuditTerms({
```
mingjiecn, line 10: Why do we need to create InternalActionAuditTerms component here? It seems the same with other types of audit, the difference is that you need to check whether the user is isAuthenticated before you display it. But you already handled it in audit-title.js.

forresttanaka: Audit title only handles the title of a facet, so if I did that, you’d still see the facet terms for internal actions, but it wouldn’t have a title — just the clickable terms. Conversely, if I only had the internal actions component but didn’t add that section to the title component, you’d get the title, but no terms under it, while logged out.

forresttanaka: Another difference is the audit title handles all audit levels, while internal actions only handles the internal actions audit.

mingjiecn (approved):

PR #496, opened 2024-05-08
IGVF-1642 Display sequencing library types on measurement set summary pages

forresttanaka:
This also fixes a bug where the Library Construction Platform doesn’t get displayed, probably because of an embedding change.


forresttanaka:
Taking it back out of code review ready. I found a long-existing bug I copied over to the new component. Quick fix, but it’ll have to go through tests again.

mingjiecn (approved):

PR #515, opened 2024-06-06
IGVF-1582-update-biosample-pages

khzlin:


components/common-data-items.js
```diff
@@ -254,6 +254,23 @@ export function SampleDataItems({
...
+                key={protocol}
+                target="_blank"
+                rel="noopener noreferrer"
+              >
+                {protocol}
+              </a>
+            ))}
+          </DataItemList>
+        </>
+      )}
```
forresttanaka, line 273: If you have more than one, do the URLs end up without separators? Like `https://www.protocols.io/link1https://www.protocols.io/link2https://www.protcols.io/link3`? I feel like this needs SeparatedList.

components/common-data-items.js
```diff
@@ -319,10 +336,10 @@ export function BiosampleDataItems({
           </DataItemValue>
         </>
       )}
-      {classification && (
+      {classifications.length > 0 && (
```
forresttanaka, line 339: If `classifications` is null, it’ll crash when you do `classifications.length`. You can fix that with `classifications?.length > 0`.

components/common-data-items.js
```diff
@@ -400,8 +427,8 @@ export function BiosampleDataItems({
 BiosampleDataItems.propTypes = {
   // Object derived from the biosample.json schema
   item: PropTypes.object.isRequired,
-  // Classification if this biosample has one
-  classification: PropTypes.string,
+  // Classifications if this biosample has one
```
forresttanaka, line 430: “has at least one”

forresttanaka (changes requested):
Looks nice. One, but probably two, changes needed.

components/common-data-items.js
khzlin: even if it's a required property?

components/common-data-items.js
khzlin: done

components/common-data-items.js
forresttanaka: Is it required? It should say that in the propTypes, and it shouldn’t have a default value on line 311 above.

khzlin: ahh i see it now

components/common-data-items.js
khzlin: done

components/common-data-items.js
```diff
@@ -257,18 +257,15 @@ export function SampleDataItems({
...
-                rel="noopener noreferrer"
-              >
-                {protocol}
-              </a>
-            ))}
-          </DataItemList>
+          <DataItemValue>
+            <SeparatedList>
+              {item.protocols.map((protocol) => (
+                <a href={protocol} key={protocol}>
```
forresttanaka, line 263: To be consistent, open the protocol page in a new tab, so you can use the `target` and `rel` attributes you had.

forresttanaka (changes requested):
Just one change.

components/common-data-items.js
khzlin: done

forresttanaka (dismissed):

pages/in-vitro-systems/[id].js
```diff
@@ -130,6 +122,22 @@ export default function InVitroSystem({
                     </DataItemValue>
                   </>
                 )}
+                {truthyOrZero(inVitroSystem.growth_medium) && (
```
Comment (forresttanaka) [inline pages/in-vitro-systems/[id].js:125]: `truthyOrZero` is for numeric values, but `growth_medium` is a string. You can simply test for its existence.

forresttanaka (changes requested):
One small change.

pages/in-vitro-systems/[id].js
Comment (khzlin) [reply pages/in-vitro-systems/[id].js]: done

PR #516, opened 2024-06-07
IGVF-1672 Update file-set tables on AnalysisSet pages

forresttanaka:
I added a fix for the dark-mode About navigation icon because I forgot about dark mode before.


mingjiecn (dismissed):
One issue is the list in Applied To Samples column can be really long and I wonder if you should make the list collapsible?

forresttanaka:
> One issue is the list in Applied To Samples column can be really long and I wonder if you should make the list collapsible?

Yeah, and I do find cases still of really long lists. I’ll write a ticket to collect them, and work on them when it seems like we have enough.

PR #530, opened 2024-06-17
IGVF-1585-update-file-pages

khzlin:


components/common-data-items.js
```diff
@@ -581,6 +581,20 @@ export function FileDataItems({ item, fileSet = null, children }) {
           <DataItemValue>{dataSize(item.file_size)}</DataItemValue>
         </>
       )}
+      {item.anvil_url && (
+        <>
+          <DataItemLabel>AnVIL Url</DataItemLabel>
+          <DataItemValue>
+            <Link href={item.anvil_url}>{item.anvil_url}</Link>
```
forresttanaka, line 588: Use an `<a>` tag with `target="_blank"` and `rel="noopener noreferrer"`. Looks like this has to be an external link, so just use the simpler `<a>` tag instead of the more complex `<Link>` tag that’s for internal links.

forresttanaka (changes requested):
One small change.

forresttanaka (dismissed):

pages/tabular-files/[...id].js
```diff
@@ -134,7 +134,7 @@ TabularFile.propTypes = {
   // Set of documents for file specifications
   fileFormatSpecifications: PropTypes.array.isRequired,
   // ConstructLibraryset this file was integrated in
-  integratedIn: PropTypes.arrayOf(PropTypes.object).isRequired,
+  integratedIn: PropTypes.array.isRequired,
```
Comment (forresttanaka) [inline pages/tabular-files/[...id].js:137]: You can still do `arrayOf(PropTypes.object)` because it’ll either be an array of objects, or an empty array, which is still a (potential) array of objects. React doesn’t have a problem with that. But I’ll leave it up to you whether you want to change that or not.

forresttanaka (approved):

pages/tabular-files/[...id].js
Comment (khzlin) [reply pages/tabular-files/[...id].js]: i can change it

forresttanaka (approved):

PR #534, opened 2024-06-20
IGVF-1741-move-socials-right-corner

khzlin:


mingjiecn (dismissed):

components/navigation.js
```diff
@@ -840,6 +847,9 @@ function NavigationCollapsed({ navigationClick, toggleNavCollapsed }) {
       <NavigationItem>
         <IndexerState isCollapsed />
       </NavigationItem>
+      <NavigationItem>
+        <Email isCollapsed />
```
forresttanaka, line 851: The `Email` component doesn’t have a property named `isCollapsed`. It would normally change the styling for the collapsed menu. Existing ones make the icon bigger for the collapsed menu, thought that might not be needed here. But in any case, you’re passing a property `isCollapsed` to a component that doesn’t accept any properties.

forresttanaka (changes requested):
The Cypress tests don’t pass, but I think it’s because this branch is nine commits behind dev. igvfd has a change that breaks Cypress tests unless you have a merge to dev from July 3.

components/navigation.js
khzlin: I removed `isCollapsed` and rebased the branch to dev.

forresttanaka (approved):

PR #536, opened 2024-06-25
IGVF-1663-samples-protocols-measurment-page

khzlin:


pages/measurement-sets/[id].js
```diff
@@ -169,6 +153,29 @@ export default function MeasurementSet({
       : [];
   const uniqueSampleSummaries = [...new Set(sampleSummaries)];
 
+  // Collect all measurement set protocols
+  let measurementSetProtocols = [];
+  if (measurementSet.protocols && measurementSet.protocols.length > 0) {
+    measurementSetProtocols = measurementSet.protocols;
+  }
```
Comment (forresttanaka) [inline pages/measurement-sets/[id].js:160]: I think this is easier to read as:
```
const measurementSetProtocols = measurementSet.protocols || [];
```
Or not even have this at all if you use the suggestion below.

pages/measurement-sets/[id].js
```diff
@@ -169,6 +153,29 @@ export default function MeasurementSet({
...
+
+  // Collect all sample protocols.
+  const sampleProtocols = [];
+  if (samples.length > 0) {
+    samples.forEach((sample) => {
+      if (sample.protocols && sample.protocols.length > 0) {
+        sampleProtocols.push(...sample.protocols);
+      }
+    });
+  }
```
Comment (forresttanaka) [inline pages/measurement-sets/[id].js:170]: This can be simpler with:
```
const sampleProtocols = samples.flatMap(sample => sample.protocols || []);
```

pages/measurement-sets/[id].js
```diff
@@ -169,6 +153,29 @@ export default function MeasurementSet({
...
+  const uniqueSampleProtocols = [...new Set(sampleProtocols)];
+
+  // Combine measurement set and sample protocols.
+  const combinedProtocols = [].concat(
+    measurementSetProtocols,
+    uniqueSampleProtocols
+  );
```
Comment (forresttanaka) [inline pages/measurement-sets/[id].js:177]: You can do:
```
const combinedProtocols = measurementSetProtocols.concat(uniqueSampleProtocols);
```
Or you could even not use `measurementSetProtocols` with:
```
const combinedProtocols = uniqueSampleProtocols.concat(measurementSet.protocols || []);
```
Is there any chance `measurementSet.protocols` and `uniqueSampleProtocols` could include the same URLs? I’m wondering if you need to deduplicate the combined list instead of deduplicating the samples list.

forresttanaka (changes requested):

pages/measurement-sets/[id].js
Comment (khzlin) [reply pages/measurement-sets/[id].js]: i removed this

pages/measurement-sets/[id].js
Comment (khzlin) [reply pages/measurement-sets/[id].js]: done

pages/measurement-sets/[id].js
Comment (khzlin) [reply pages/measurement-sets/[id].js:163]: i deduplicated the combined list and removed the code for deduplicating the samples list.

forresttanaka (approved):

PR #537, opened 2024-06-25
IGVF-1647 Scrollable report-view column-selection modal

forresttanaka:
While this ticket ostensibly has to do with working around a Safari bug, I decided to take this chance to improve the UI so that very tall column-selection modals (like you’d get with `/multireport/?type=Item`) don’t become awkwardly tall. I have defined a maximum and minimum height for the modal that adjusts itself to the height of the browser window.

I renamed the existing scroll-indicator.js used on the report-view table to grid-scroll-indicators.js to distinguish it from the new vertical-scroll-indicator.js meant for any scrollable area to indicate that the area is in fact scrollable. I might update grid-scroll-indicators.js in the future to share more with scroll-indicator.js, but that will be for another day.

### DIFF
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -6,6 +6,16 @@ module.exports = {
       center: true,
     },
     extend: {
+      animation: {
+        "scroll-fade": "scrollFade 3s linear",
+      },
+      // that is actual animation
+      keyframes: () => ({
+        scrollFade: {
+          "0%": { opacity: 1 },
+          "100%": { opacity: 0 },
+        },
+      }),
       aspectRatio: {
         hd: "16 / 9",
         cinema: "21 / 9",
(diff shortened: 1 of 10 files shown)


forresttanaka:
To be clear, I’m aiming this ticket for next week’s release, not today’s.

tailwind.config.js
```diff
@@ -250,6 +260,9 @@ module.exports = {
         "min-2": "repeat(2, minmax(0, min-content))",
         "data-item": "fit-content(200px) 1fr",
       },
+      maxHeight: {
+        "column-select-modal": "calc(100vh - 300px)",
```
mingjiecn, line 264: very cool!

mingjiecn (approved):

tailwind.config.js
forresttanaka: Yeah!

PR #547, opened 2024-07-05
IGVF-1746 Add batch-download button to file tables and file-set searches

forresttanaka:
I guess Jest changed at some point, and now many Jest tests with interactions fail because they require an `act()` wrapper. So that explains many Jest test changes.

Discovered haven’t used the instruction.js module in a while, so I deleted that and its Jest test.

I changed the prose-based styles in globals.css because I realized I wanted to use the prose styles not only on pages, but anywhere. So instead of targeting `#page-content` I instead target `.prose`. I don’t also have to target `dark:prose-invert` because I always use both classes together, so targeting one targets the other.

I removed the auto-focus mechanism when a modal gets displayed initially because it doesn’t work with the latest headlessui version. You instead have to use the `autoFocus` or `data-autofocus` attributes.

I added a `downloadQuery` property to `FileTable` but I don’t use it for anything. I think once we have more embedding for FileSet objects in igvfd, we’ll need this.


forresttanaka:
I added markdown documentation for the new batch-download mechanism, but that .md file is the only change.

mingjiecn (approved):

PR #559, opened 2024-07-22
IGVF-1819 Use status=released on links to list and report views

forresttanaka:
I had to remove the h1 tests in the app.cy.js Cypress test because with the `status=released` in the query string, some types show no results, so this test has to just make sure the page comes up without that test.

I have a couple other changes in this branch caused by a Cypress change that affects Typescript.


mingjiecn (approved):

forresttanaka:
Sorry; you code reviewed small changes for this branch before, and it has grown to a lot of changes. Mostly this comes from redesigning the breadcrumb system. I hadn’t imagined breadcrumbs needing browser states (in this case, user logged in as admin), so my original design of doing everything on the NextJS server and passing that in through props doesn’t work. So now the `<Breadcrumb>` component calls `buildBreadcrumbs()` instead of each page calling that on the server. If breadcrumbs need something from the server, use the new breadcrumb meta. Right now only Page breadcrumbs use meta from the server. The Items breadcrumbs also use meta, but to transfer the browser `isAdmin` state to `buildBreadcrumbs()`.

Most of the changes to the pages involve removing breadcrumb generation, and passing the page’s object to `<Breadcrumbs>`.

I changed `<Attribution>`  because I saw some cases where the Attribution box could appear empty.

mingjiecn (approved):
My question is why we have to use different status filter for admin and non_admin?

forresttanaka:
@mingjiecn 
> why we have to use different status filter for admin and non_admin?

Wranglers didn’t want the pink status!=deleted button on the display for submitters, only wranglers. It’s OK for submitters to see deleted objects. And it’s OK for wranglers too, but the deleted objects were annoying for wranglers, as they normally sort at the head of the list. Sorting list views differently was one option, but this was the preferred option in the end.

In a sense, we don’t have any good options, just less bad options. This was triggered (if I understood correctly) by a huge number of objects getting deleted and resubmitted.

image

mingjiecn:
> @mingjiecn
> 
> > why we have to use different status filter for admin and non_admin?
> 
> Wranglers didn’t want the pink status!=deleted button on the display for submitters, only wranglers. It’s OK for submitters to see deleted objects. And it’s OK for wranglers too, but the deleted objects were annoying for wranglers, as they normally sort at the head of the list. Sorting list views differently was one option, but this was the preferred option in the end.
> 
> In a sense, we don’t have any good options, just less bad options. This was triggered (if I understood correctly) by a huge number of objects getting deleted and resubmitted.
> 
> image

OK. Got it. Thanks for the explanation!

PR #560, opened 2024-07-22
IGVF-1755-tabular-table-aux-page

khzlin:


forresttanaka (approved):
Nice! BTW I mentioned in the meeting about upgrading to NextJS 14 at some point (we currently use NextJS 13) I think I’ll need to convert the entire igvf-ui to Typescript at that point.

PR #584, opened 2024-08-13
IGVF-1849 ORF genes property update

forresttanaka:
The Cypress test fails because it needs IGVF-1792 to get merged into igvfd dev.


pedrohr (dismissed):

pedrohr:
The code looks good! The breaking cypress test seems to be unrelated.

PR #601, opened 2024-08-28
IGVF-1895 Fix generic page crash

forresttanaka:
I’d forgotten that generic pages and report pages shared a bit of code.

### DIFF
--- a/components/unknown-type-panel.js
+++ b/components/unknown-type-panel.js
@@ -56,15 +56,23 @@ export default function UnknownTypePanel({ item, attribution = null }) {
         <DataArea>
           <CommonDataRenderer item={item}>
             {_.sortBy(genericProperties).map((property) => {
-              const label = profile?.properties[property]?.title || property;
-              return (
-                <Fragment key={property}>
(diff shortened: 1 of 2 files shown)


components/unknown-type-panel.js
```diff
@@ -56,15 +56,23 @@ export default function UnknownTypePanel({ item, attribution = null }) {
...
-              const label = profile?.properties[property]?.title || property;
-              return (
-                <Fragment key={property}>
-                  <DataItemLabel>{label}</DataItemLabel>
-                  <DataItemValue>
-                    <UnspecifiedProperty property={item[property]} />
-                  </DataItemValue>
-                </Fragment>
-              );
+              if (!Array.isArray(item[property])) {
```
mingjiecn, line 59: Does this mean we don't display the property if it is an array?

mingjiecn (approved):
Just have one question. Thanks!

components/unknown-type-panel.js
forresttanaka: Yes. The generic pages are just to get an idea of what's in the object.

PR #629, opened 2024-10-07
IGVF-1976-varchamp-collection-icon

khzlin:


forresttanaka (changes requested):
Is the SVG based on their prior art, or you came up with it? If it’s yours, I’d suggest kerning the “Va” closer together because it kind of looks like “V arCHAMP.” Sometimes “Optical” kerning Illustrator’s text layer’s properties is enough, but I often have to do it manually.

Should “across” be capitalized? I realize their GitHub repo doesn’t nor does their grant, and we probably have to stay consistent with them, but it seems like a mistake on their part.

I have rules about the badge that I kept on my local drive but never published, and should have: 210px × 140px (which you have), maximum content height: 120px, maximum content width: 190px. The width of the “VarChAMP” part is 204px which puts it really close to the edge.

khzlin:
@forresttanaka The svg is based on the prior art. I thought I should make it consistent with what was done before. Thanks for the rules, I'll fix the image.

forresttanaka (changes requested):
Sorry, I’m now noticing the new “VarChAMP” is centered in the rectangle correctly, but the small words are shifted to the right. Also, the green background’s left edge doesn’t line up with the left edge of the view box, leaving a small white gap.

forresttanaka (dismissed):

PR #632, opened 2024-10-14
IGVF-1919-display-properties

khzlin:


pages/analysis-sets/[id].js
```diff
@@ -82,6 +83,31 @@ export default function AnalysisSet({
               )}
               <DataItemLabel>File Set Type</DataItemLabel>
               <DataItemValue>{analysisSet.file_set_type}</DataItemValue>
+              {analysisSet.summary && (
+                <>
+                  <DataItemLabel>Summary</DataItemLabel>
+                  <DataItemValue>{analysisSet.summary}</DataItemValue>
+                </>
+              )}
+              {analysisSet.functional_assay_mechanisms?.length > 0 && (
```
Comment (forresttanaka) [inline pages/analysis-sets/[id].js:92]: I don’t _think_ you need the question mark. It looks like `functional_assay_mechanisms` always exists, though it could be empty.

pages/measurement-sets/[id].js
```diff
@@ -228,6 +228,25 @@ export default function MeasurementSet({
                     </DataItemList>
                   </>
                 )}
+                {measurementSet.functional_assay_mechanisms?.length > 0 && (
```
Comment (forresttanaka) [inline pages/measurement-sets/[id].js:231]: Similar comment to analysis set.

forresttanaka (changes requested):
Just a single one-character change. Nice you were able to integrate the panel collapsing.

pages/analysis-sets/[id].js
Comment (khzlin) [reply pages/analysis-sets/[id].js]: removed.

pages/measurement-sets/[id].js
Comment (khzlin) [reply pages/measurement-sets/[id].js]: removed

forresttanaka (approved):

PR #633, opened 2024-10-14
IGVF-1954 File graphs on file-set pages

forresttanaka:
I added a partial documentation page because the d3-dag package is so poorly documented, and I wanted to be sure I didn’t forget how to use some parts of it. Also, I’ve started to remove the types from the TSDoc function-parameter documentation, as VSCode doesn’t use it to color TSDoc correctly, nor does VSCode show the types in Intellisense. You can see the types just a few lines down anyway.


lib/files.ts
```diff
@@ -59,10 +70,116 @@ export function checkForFileDownloadPath(path: string): boolean {
...
+      processedFileIds.add(file["@id"]);
+    });
+
+    // We now have a batch of files that might derive from other files. Request any derived from
+    // files in this batch, which gets us a new upstream batch of files to process.
+    derivedFromFileBatch = await getFileDerivedFromFiles(
+      derivedFromFileBatch,
+      request,
+      processedFileIds
+    );
```
pedrohr, line 176: is there a situation where the recursion level here can go too deep? not sure if it makes sense to add a limit to the recursion level

forresttanaka: I can’t _think_ of a situation where it could go too deep. It has detected loops I’ve set up, including multi-node loops. And I’m not sure how deep we can practically expect these to go right now, so if we chose a depth, I have no recommendation what that depth should be.

pedrohr: sounds good, we can adapt in the future if we see any complicated cases to draw on the screen

pedrohr (approved):

PR #653, opened 2024-11-13
IGVF-2147 Fix Cypress indexing delays

forresttanaka:


cypress/e2e/access-keys.cy.js
```diff
@@ -22,7 +22,7 @@ describe("Exercise access keys", () => {
 
     // Reset access key
     cy.get(`[aria-label^="Reset access key"]`).click();
-    cy.delayForIndexing();
+    cy.wait(2000);
```
keenangraham, line 25: Does this need to go back to `delayForIndexing`?

keenangraham (approved):

PR #683, opened 2025-01-06
IGVF-2001 Home page update

forresttanaka:


forresttanaka:
> I tried this api

Yeah that’s odd. I’ll investigate, though it might take some time unless I can reproduce something like that locally.

forresttanaka:
@mingjiecn That URL now returns data. I think maybe you used that URL which, when passed to igvfd really returns no data (https://igvfd-igvf-2001-home-updates.demo.igvf.org/dataset-summary-agg/?type=MeasurementSet). This lack of data then got cached. The cache expiry time is 15 minutes, so once that expired, it started returning data.

The correct URL is https://igvf-ui-igvf-2001-home-updates.demo.igvf.org/api/dataset-summary/?type=MeasurementSet&config=PreferredAssayTitleSummary

mingjiecn (dismissed):
Cache is tricky!

PR #689, opened 2025-01-17
IGVF-2143 Remove facet groups and display all facets

forresttanaka:


components/facets/facet-section.js
```diff
@@ -49,60 +48,23 @@ ClearAll.propTypes = {
  * facet-group button, display the modal containing the facets for that group.
```
mingjiecn, line 48: need to change the documentation here since we don't use facet-group anymore?

forresttanaka: Yes.

mingjiecn (dismissed):

PR #690, opened 2025-01-21
IGVF-2247-analysis-step-version-table

khzlin:


components/analysis-step-version-table.js
```diff
@@ -0,0 +1,78 @@
...
+          ))}
+        </SeparatedList>
+      );
+    },
+    isSortable: false,
+  },
+];
+
+/**
+ * Display a sortable table of the given analysis steps. Optionally display a link to a report page of
```
forresttanaka, line 35: really for the given analysis step *versions*.

forresttanaka (changes requested):
Just a comment correction. One thing I haven’t yet told you is that I’d like to stop using default exports. It causes confusion, and even those familiar have to keep looking up whether something is default exported or not so they know whether to `import { a } from "a";` or `import a from "a";`. I think it’s a lot easier to always do the first and not ever have to do the second.

components/analysis-step-version-table.js
khzlin: fixed

forresttanaka (approved):

forresttanaka (approved):

PR #702, opened 2025-02-06
IGVF-2264 Tissue summary page

forresttanaka:
The most important file here is `pages/tissue-summary/[taxa].tsx` as it has all the new functionality. I had to update the data grid component a bit to work with it and to allow better `role` use.

### DIFF
--- a/pages/tissue-summary/[taxa].tsx
+++ b/pages/tissue-summary/[taxa].tsx
@@ -426,7 +426,7 @@ export default function TissueSummary({
       <div className="pt-4">
         <LabelXAxis label={tissueSummary.matrix.x.label} />
         <div className="flex">
-          <LabelYAxis label={tissueSummary.matrix.y.label} />
+          <LabelYAxis label="Tissue / Organ" />
           <div
             role="table"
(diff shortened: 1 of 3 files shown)


pages/tissue-summary/[taxa].tsx
```diff
@@ -0,0 +1,490 @@
...
+ */
+function RowSubheaderCell({
+  source,
+  meta,
+}: {
+  source: CellSource;
+  meta: { taxa: string };
+}) {
+  const bgClass = taxaQueries[meta.taxa].rowSubheaderCellClass;
+  const hoverClass = taxaQueries[meta.taxa].hoverCellClass;
```
Comment (mingjiecn) [inline pages/tissue-summary/[taxa].tsx:158]: Is it possible to add hoverClass to RowHeaderCell and ColumnHeaderCell? Nobody wants it?

mingjiecn (dismissed):

pages/tissue-summary/[taxa].tsx
Comment (forresttanaka) [reply pages/tissue-summary/[taxa].tsx:158]: People said that the row header cell hover was optional. `RowHeaderCell` can span multiple rows, so that would have been difficult to do without either making a custom version of `DataGrid` specifically for this page, or adding this generically in `DataGrid` which will take some effort I think.

I’ll need to do some generic modifications to `DataGrid` for future matrices (like sticky headers) so this ability might happen then, or possibly because it wasn’t seen as important, whereas the sub-row hover was seen as very important.

PR #713, opened 2025-02-20
IGVF-2332 Cell Model Page

forresttanaka:


cdk/infrastructure/config.py
```diff
@@ -66,6 +66,7 @@
                 'max_capacity': 4,
                 'use_redis_named': 'Redis71',
             },
+            'backend_url': 'https://igvfd-igvf-2354-cell-model-matrix-config.demo.igvf.org',
```
pedrohr, line 69: is this a leftover from a test? this config is not being used in this PR

forresttanaka: Yes, because the igvfd branch that this branch requires is still in code review, so I have to point this demo at the demo for that branch. Once I merge this (or if that igvfd branch gets merged to dev), then I remove this line.

forresttanaka: That’s also why this PR has a failing test: it tests to make sure this line doesn’t exist, which it won’t once this gets merged.

pedrohr: oh I see! I wasn't aware! All good 👍

pedrohr (dismissed):

PR #719, opened 2025-02-28
IGVF-2353 Institutional certificates UI improvements

forresttanaka:


forresttanaka:
@mingjiecn I think the igvfd branch that this branch relies on won’t get merged today, so I’m going to change the CDK of this branch to point at the igvfd demo I need. So if you see a change to this branch in a bit, it’s just the CDK change.

components/data-area.d.ts
```diff
@@ -88,7 +88,7 @@ export function DataItemValueUrl({
   children,
 }: {
   className?: string; // Additional Tailwind CSS classes
-  children: string; // URL to display
+  children: React.ReactNode; // URL to display
```
mingjiecn, line 91: I thought url should always be a string? Why do we need to change the type from string to React.ReactNode

forresttanaka: It used to always be a string, but now I pass things like links as children, which doesn’t work for the string type.

mingjiecn: oh, OK. thanks!

mingjiecn (approved):

PR #738, opened 2025-03-25
IGVF-2360 View object history

forresttanaka:


components/history-link.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+  if (item && basePath) {
+    throw new Error("You must provide either an item or a basePath, not both.");
+  }
+
+  // Include labels to go to history view of the object or the object itself?
+  const isHistoryView = Boolean(item);
+
+  // Must be logged in as admin for this to work. Check both for weird cases where one is null.
+  if (isAuthenticated && sessionProperties?.admin) {
```
mingjiecn, line 65: should also add the else case to return null?

forresttanaka: I used to, and I think old versions of React required it, but now React is fine rendering nothing if you return undefined, so that’s what I usually do now.

mingjiecn (approved):

PR #748, opened 2025-04-03
IGVF-2571-resource-toolbar

khzlin:


components/navigation.js
```diff
@@ -681,62 +681,13 @@ function NavigationExpanded({ navigationClick, toggleNavCollapsed }) {
...
-
-        <NavigationGroupItem
-          id="resources"
-          title="Resources"
-          icon={<Icon.Resources />}
-          isGroupOpened={openedParents.includes("resources")}
-          handleGroupClick={handleParentClick}
-        >
           <NavigationHrefItem
             id="overview"
```
forresttanaka, line 685: Change the `id` to `resources`. It doesn’t make a technical difference, but might as well keep things consistent.

forresttanaka (changes requested):

forresttanaka (approved):

PR #753, opened 2025-04-04
IGVF-2578 Barcode map fixes

forresttanaka:
`.unwrap_or({})` returns the requested object if successful, or an empty object if not. The React code was expecting a null instead of an empty object. This replaces that with `.optional()` so failure returns a null.


khzlin (approved):
this looks good to me

PR #756, opened 2025-04-11
IGVF-2499-curated-set-file-table

khzlin:


components/curated-file-table.js
```diff
@@ -0,0 +1,148 @@
...
+    id: "upload_status",
+    title: "Upload Status",
+    display: ({ source }) => <Status status={source.upload_status} />,
+  },
+];
+
+/**
+ * Display a sortable table of the given files.
+ */
+export default function CuratedFileTable({
```
forresttanaka, line 66: Could you make this a non-default export, and then import it with `import { CuratedFileTable } from "../../components/curated-file-table";`? I’d like to make all new components non-default exports.

forresttanaka (changes requested):
When I first saw this, I thought there might be a new CuratedFile object type. Would it be better to name the file `curated-set-file-table.js` and the component `CuratedSetFileTable`. I’ll leave that up to you, but I was confused for a bit.

forresttanaka:
I considered suggesting _not_ making a new component, but to make a conditional `submitted_file_name` column in the existing `FileTable` component to reduce code duplication. But you have also removed the “Filtered” column in this new component, which I think tips the scale towards a new component like you have. But it’s something to consider in future tables.

forresttanaka (dismissed):

forresttanaka (approved):

PR #757, opened 2025-04-11
IGVF-2517 Sample term descriptions and definitions

forresttanaka:


components/common-data-items.js
```diff
@@ -509,6 +517,14 @@ export function OntologyTermDataItems({ item, isA, children }) {
           </DataItemValue>
         </>
       )}
+      {(item.description || item.definition) && (
```
mingjiecn, line 520: I think term needs to show definition if there is one. if not then show description if there is one. Description may not be as accurate as definition.

mingjiecn (changes requested):

components/common-data-items.js
forresttanaka: Oh really? I guess Cricket has it reversed then.

mingjiecn: I am positive definition is more accurate.

mingjiecn (approved):

PR #771, opened 2025-04-29
IGVF-2633-standardized-document-column

khzlin:


components/document-table.js
```diff
@@ -40,6 +41,15 @@ const columns = [
...
   },
+  {
+    id: "standardized_file_format",
+    title: "Standardized File Format",
+    display: ({ source }) => {
+      return source.standardized_file_format ? (
+        <Status status="standardized" />
+      ) : null;
+    },
+  },
```
forresttanaka, line 52: Does this column sort the way you want? When you sort ascending, all the ones with `standardized_file_format` true sort at the end.

forresttanaka (approved):
One question on column sorting, otherwise it’s good.

components/document-table.js
forresttanaka: Should it simply be non-sortable?

PR #776, opened 2025-05-02
IGVF-2481-step-version-model-page

khzlin:


pages/analysis-step-versions/[id].js
```diff
@@ -124,7 +124,9 @@ export async function getServerSideProps({ params, req, query }) {
     return {
       props: {
         analysisStepVersion,
-        pageContext: { title: analysisStepVersion.uuid },
+        pageContext: {
+          title: `${analysisStepVersion.analysis_step.title} ${"Version"}`,
```
Comment (forresttanaka) [inline pages/analysis-step-versions/[id].js:128]: Since “Version” is just a static string, you don’t need it in a `${}`. Just `${analysisStepVersion.analysis_step.title} Version`

pages/model-sets/[id].js
```diff
@@ -243,9 +244,6 @@ export async function getServerSideProps({ params, req, query }) {
     await request.getObject(`/model-sets/${params.id}/`)
   ).union();
   if (FetchRequest.isResponseSuccess(modelSet)) {
-    const softwareVersion = modelSet.software_version
-      ? (await request.getObject(modelSet.software_version)).optional()
-      : null;
     const documents = modelSet.documents
```
Comment (forresttanaka) [inline pages/model-sets/[id].js:247]: Delete the `softwareVersion` in the returned props of this function. It’ll probably crash when you visit a ModelSet page.

forresttanaka (changes requested):

pages/model-sets/[id].js
Comment (khzlin) [reply pages/model-sets/[id].js:247]: ok

pages/analysis-step-versions/[id].js
Comment (khzlin) [reply pages/analysis-step-versions/[id].js:128]: ok

forresttanaka (approved):

PR #781, opened 2025-05-07
IGVF-2195 Date-range facet

forresttanaka:


components/facets/custom-facets/date-range-lib.ts
```diff
@@ -0,0 +1,171 @@
...
+ * format "YYYY-MM-DD". For the "gt:" case, the date gets adjusted one day later. For the "lt:"
+ * case, the date gets adjusted one day earlier. If the term is invalid, the function returns an
+ * empty string. Generally, this function gets called after we already know the term is in one of
+ * the valid formats, not like a wildcard or a single date.
+ * @param term - gte:, lte:, gt:, or lt: term to extract the date from
+ * @returns date in the format YYYY-MM-DD; empty string if the term is invalid
+ */
+export function extractDateFromFilterTerm(term: string): string {
+  // Check for "gte:YYYY-MM-DD" or "lte:YYYY-MM-DD" format.
+  const matchOrEqual = term.match(/(?:gte|lte):(.+)/);
```
pedrohr, line 22: should we replace by something like (\d{4})-(\d{2})-(\d{2}) or any string after the lte|gte: is fine?

components/facets/custom-facets/date-range-lib.ts
```diff
@@ -0,0 +1,171 @@
...
+    const date = stringToDate(iso8601Date);
+    if (!isNaN(date.getTime())) {
+      return iso8601ToDateOnly(iso8601Date);
+    }
+    return "";
+  }
+
+  // Check for "gt:YYYY-MM-DD" or "lt:YYYY-MM-DD" format. The UI doesn't make use of these
+  // filters, but we need to handle them in case the user has added them manually.
+  const match = term.match(/(?:gt|lt):(.+)/);
```
pedrohr, line 34: same question as above

components/facets/custom-facets/date-range-lib.ts
forresttanaka: I’m relying on the back end to filter out the weirder things, for example:
```
https://igvf-ui-igvf-2195-date-range-facet.demo.igvf.org/search/?type=MeasurementSet&status!=deleted&release_timestamp=gt:202a
```
causes a 503 error, which is fine to me.

You _can_ enter things the back end does take and parse, like
```
https://igvf-ui-igvf-2195-date-range-facet.demo.igvf.org/multireport/?type=MeasurementSet&status!=deleted&release_timestamp=gte:2024
```
…which returns anything in 2024 or after. This UI ignores this because the function `stringToDate()` expects yyyy-mm-dd, and returns an invalid date with just `2024`. So I _think_ that achieves the same as a more restrictive regex like you have; just a little bit later.

I could see the argument that the UI should interpret things like `gte:2024` the same way as the back end (though sort of oddly, given `gt:2024` the back end also returns items in 2024, not just 2025 and beyond; I think it probably treats `2024` as `2024-01-01` and would return dates starting `2024-01-02`. But I felt like ignoring things that don’t follow the pattern was simpler instead of trying to copy the functionality of the back end.

pedrohr (approved):

PR #819, opened 2025-06-16
IGVF-2816-display-summary

khzlin:


forresttanaka:
Interesting. I think `summary` is calculated though, right?

PR #825, opened 2025-06-20
IGVF-2736 Remove workflow from analysis-step pages

forresttanaka:
Remove workflow display from analysis-step page.


forresttanaka:
> ## Pull Request Overview
> * Since the workflow block is removed from the UI, consider updating the data fetching logic (e.g., GraphQL query or API endpoint) to stop retrieving the `workflow` field and avoid unused data loads.

I don’t fetch `workflow` from `getServerSideProps()`.

> * Add or update unit/integration tests to verify that the workflow section is no longer rendered and that the layout remains correct when workflow data is omitted.

We don't have Jest nor Cypress tests to test individual properties displayed on object pages.

PR #841, opened 2025-07-10
IGVF-2843 Reduce CircleCI Cypress Errors

forresttanaka:


forresttanaka:
Making the tests more reliable involved changing the Cypress memory management while running under CircleCI, as well as suppressing the harmless `performance` errors that would still cause the entire test to fail. Other changes were to make Cypress test scripts compatible with the current ESLint version. The add.cy.js test used to comply with ESLint, but changes to ESLint caused a lot of conflicts in that file.

mingjiecn (approved):

PR #842, opened 2025-07-10
IGVF-2792 Fix section directory for JSON display

forresttanaka:
You can trigger the rebuilding of the section directory by changing the hash you pass to it. I now pass a new property to `useSecDir` — `isJson` — that changes the hash when switching between normal and JSON modes, triggering the rebuilding of the section directory so that it only shows **Top of Page** while in JSON mode, and then showing all the section on the page while in normal mode.

I also updated the highlighting when you choose a section directory item and scroll to the corresponding section.

I merged in changes from the IGVF-2824-optimize-tailwind branch as they’re just too small to justify their own ticket. This cleans up the Tailwind CSS configuration to better work with Tailwind CSS v4.

### DIFF
--- a/styles/globals.css
+++ b/styles/globals.css
@@ -678,22 +678,45 @@
 }
 
 /**
- * For highlighting the a section directory anchor when the user selects it. It appears with a
- * background color, then fading to transparent.
+ * For highlighting the a section directory anchor when the user selects it. This shows a flashing
+ * dot to the left of the section directory anchor.
  */
-.bg-section-directory {
-  animation: highlight 2s linear;
-}
+@layer utilities {
+  @keyframes flash-multi {
+    0%,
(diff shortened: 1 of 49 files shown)


forresttanaka:
> Because isJson defaults to false, typeof isJson === "boolean" is always true and will always override any hash value

Yes, that’s true. I’ll remove the default.

styles/globals.css
```diff
@@ -678,22 +678,45 @@
 }
 
 /**
- * For highlighting the a section directory anchor when the user selects it. It appears with a
- * background color, then fading to transparent.
+ * For highlighting the a section directory anchor when the user selects it. This shows a flashing
```
mingjiecn, line 681: I like the old highlighting method though...

PR #861, opened 2025-08-04
IGVF-2902-cls-samples-donors

khzlin:


forresttanaka (dismissed):
Nice, using the new `Object.property?.length > 0` mechanism. Hopefully that makes everything easier and more reliable.

PR #1055, opened 2026-05-26
merge-dev-to-main-8.97.0

ottojolanki:


forresttanaka (approved):
This the same PR as before? I thought I approved it a while ago. Not sure what happened.

ottojolanki:
> This the same PR as before? I thought I approved it a while ago. Not sure what happened.

Yeah, there ever was only one. Wouldn't be surprised if github sometimes just dropped requests..

PR #1068, opened 2026-06-16
IGVF-3557 Update construct library set ORF displays

khzlin:


pages/construct-library-sets/[id].js
```diff
@@ -122,12 +122,12 @@ function LibraryDetails({ library }) {
               </DataItemValue>
             </>
           )}
-          {library.orf_list && (
+          {library.small_scale_orf_list && (
             <>
-              <DataItemLabel>Open Reading Frame List</DataItemLabel>
+              <DataItemLabel>Smalle Scale ORF List</DataItemLabel>
```
Comment (forresttanaka) [inline pages/construct-library-sets/[id].js]: Misspelling — “Ye Smalle Haberdashery.” Technically “small scale” is a noun phrase modifying “ORF” so it should — technically — be “Small-Scale ORF.” It’s an ORF that's small scale, not a scale ORF that's small. I know biological terms can seem weird when perfectly grammatically correct though, so I’ll leave that one up to you.

pages/construct-library-sets/[id].js
```diff
@@ -136,6 +136,19 @@ function LibraryDetails({ library }) {
               </DataItemValue>
             </>
           )}
+          {library.large_scale_orf_list && (
+            <>
+              <DataItemLabel>Large Scale ORF List</DataItemLabel>
+              <DataItemValue>
+                <Link
+                  href={library.large_scale_orf_list["@id"]}
+                  key={library.large_scale_orf_list["@id"]}
```
Comment (forresttanaka) [inline pages/construct-library-sets/[id].js]: You don’t need a key. This `Link` isn’t rendered inside of a loop. Doesn’t hurt anything, but it’s just not needed.

forresttanaka (changes requested):

pages/construct-library-sets/[id].js
Comment (khzlin) [reply pages/construct-library-sets/[id].js]: done

pages/construct-library-sets/[id].js
Comment (khzlin) [reply pages/construct-library-sets/[id].js]: i'm thinking of reviewing "small scale" and "large scale" completely since it's applicable to the gene list and loci list but would like to pass it through the wranglers first. I'll create a ticket for a later sprint.

Comment (forresttanaka) [reply pages/construct-library-sets/[id].js]: OK, so it's supposed to be spelled “Smalle”?

Comment (khzlin) [reply pages/construct-library-sets/[id].js]: oh no. I switched it to "Small"

forresttanaka (approved):

PR #1098, opened 2026-08-03
IGVF-2895 Update metadata-download directions in the download modal

keenangraham:


forresttanaka (dismissed):
Looks nice! The place where you have `&apos;^#&apos;` I _think_ you can enter regular straight single quotes there. I use things like `&ldquo;` partly because GitHub displayed real directional double quotes wrong a long time ago, and now I do that to make it clear these aren’t normal double quotes. But what you have is completely correct too (again, assuming it works), so it’s up to you.

forresttanaka (approved):

